PR #4, opened 2018-04-04
Ported enum from Python 3.6.5.

Thom1729:
Importable as `sublime_lib.enum`.

I'm not sure if there's a better idiom for `enum/__init__.py` than `from .enum import *`. That does seem to exactly express the intent.

I'd like to port the tests as well, but that looks to be more complicated.


FichteFoll:
Do we need to expose the module at all? If yes, an `import *` seems to be the correct call, anyway.

Do you think we will eventually vendor other things than Python modules/packages?

FichteFoll:
Also, we could of course also add the backported `enum` as a PC dependency instead of the current `enum34`. Would still mean that users of sublime_lib need to add both to their `dependencies.json`, though.

Thom1729:
I don't know if we'll need to vendor anything else. If not, we have a superfluous directory or so, but if we do, it will be nice to have things in order. I'm open to other ways of organizing the code.

I do think it's beneficial to expose it. We're going to expose enums based on it anyway, so exposing the underlying library doesn't really add complexity, whereas it might be useful to package authors. If we later make a 3.7 version of `sublime_lib`, we can eliminate the vendor code and change the import to seamlessly refer to the native library (while maintaining compatibility with packages relying on `sublime_lib.enum`).

Using a dependency would be cleaner in some ways, but because users would have to manually depend on the other package I think it's overall less messy to bundle it. When Package Control revamps dependencies, we can have the best of both worlds.

PR #5, opened 2018-04-04
Added .flake8 config.

Thom1729:
This should get us started. I don't have a strong opinion about other rules.


FichteFoll:
I suppose the only strong opinion I have is splitting lines *before* binary operators, but the check for that hasn't been released yet (part of pycodestyle).

Issue #3, opened 2018-04-03
Output Panel Abstraction


Thom1729:
The PackageDev implementation is here: https://github.com/SublimeText/PackageDev/blob/d22a77646b3b3e9ab06f81ab0b6037ea17ef2380/plugins_/lib/sublime_lib/view/output_panel.py

Thom1729:
Comments on the existing implementation:

- Adding `flush` (a no-op) would make it file-like.
- What about renaming `write_line` to `print`?

FichteFoll:
- +1 for `flush`
- Technically, since the class is file-like, `print(file=output_panel)` could be used, but I'm not sure whether that is clear enough. If we think of it as an abstraction over a file (as it is file-like), then we should perhaps just scrape `write_line` completely and have users resort to their usual methods of adding newlines to a file or some other buffer.

FichteFoll:
We should also think of how a `LogHandler` targetting an output panel should be implemented for use with the `logging` stdlib.

Thom1729:
It should work with `StreamHandler`, at the least.

On a related note, it might be useful to implement a simple `ViewStream` that implements `write` and `flush` and have the output view subclass that. That would be handy for anyone who wants to build up a view in a file-like manner.

PR #6, opened 2018-04-04
Added ViewStream and OutputPanel.

Thom1729:
Discussed briefly in #3.

Notes:

- `write` and `flush` constitute the informal stream interface. `flush` is, of course, a no-op. (I don't think there's any reason to do any buffering here.)
- `print` is a convenience method: `stream.print(foo)` rather than `print(foo, file=stream)`. It might be appropriate to throw a `TypeError` if the user passes a `file` argument.
- The behavior of `seek_start` is used in the `finish` method of the existing `OutputPanel` implementation. Because it is purely a view operation, I thought it made sense to implement at this lower level. `seek_end` and `seek` make sense in combination with `seek_start`.
- `clear` uses two commands, which could cause flickering in some extreme cases. However, the only workaround is creating a new command, which would require a workaround of its own to get working in a dependency. I don't feel that it's worth the complication, myself.

### DIFF
--- a/st3/sublime_lib/view_stream.py
+++ b/st3/sublime_lib/view_stream.py
@@ -6,10 +6,17 @@ class ViewStream():
         self.view = view
 
     def _check_selection(self):
+        if len(self.view.sel()) == 0:
+            raise ValueError("The underlying view has no selection.")
         if len(self.view.sel()) > 1:
             raise ValueError("The underlying view has multiple selections.")
 
+    def _check_is_valid(self):
+        if not self.view.is_valid():
+            raise ValueError("The underlying view is invalid.")
+
     def write(self, s):
+        self._check_is_valid()
         self._check_selection()
         self.view.run_command('insert', { 'characters': s })
         return len(s)
@@ -21,6 +28,7 @@ class ViewStream():
         pass
 
     def seek(self, index, whence=SEEK_SET):
+        self._check_is_valid()
         if whence == SEEK_SET:
             selection = self.view.sel()
             selection.clear()
@@ -41,5 +49,6 @@ class ViewStream():
         self.seek(self.view.size())
 
     def clear(self):
+        self._check_is_valid()
         self.view.run_command('select_all')
         self.view.run_command('left_delete')
--- a/st3/sublime_lib/view_stream.py
+++ b/st3/sublime_lib/view_stream.py
@@ -5,7 +5,12 @@ class ViewStream():
     def __init__(self, view):
         self.view = view
 
+    def _check_selection(self):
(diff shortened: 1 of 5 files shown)


FichteFoll:
Since `seek_start()` is equivalent to `seek(0)` (which is actually less characters), I'd rather be inclined to add a second parameter similar to `io.IOBase.seek`. That's re-using stdlib semantics with little added complexity.
For clarification, `seek_end()` would then become `seek(0, io.SEEK_END)`.

Do you think we could use safe-guards aganinst situations with less or more than a single selection?

Thom1729:
Maybe we should do both for `seek`. In general, I prefer separate methods to flag arguments that change behavior, and I definitely think that `seek_end()` is simpler than `seek(0, io.SEEK_END)`. (The latter looks suspiciously like a C API wrapper.) However, we can augment the `seek` method to be fully compatible with the `io` convention while keeping the convenience method.

I'm not sure how multiple selection safeguards would work. Would the method throw an exception? Only if a flag argument were (not) passed? It's not obvious to me what behavior would be sensible, other than what it's currently doing.

Thom1729:
I've added a check. If you call `write`, and there are multiple selections, it will raise a `ValueError`. Is this more or less what you were thinking of? (Also, should it be an `IOError`?)

Thom1729:
I've added a basic OutputPanel, implementing `show`, `hide`, and `destroy`.

- Should `mypanel.show(False)` hide the panel? Is this useful?
- Would it be reasonable to default the `window` argument to `sublime.active_window`, or is that confusing? If that's a possibility, we should switch the order of the constructor arguments.
- Should we raise an error if you call a method of a panel that's been destroyed?

st3/sublime_lib/view_stream.py
```diff
@@ -5,7 +5,12 @@ class ViewStream():
     def __init__(self, view):
         self.view = view
 
+    def _check_selection(self):
+        if len(self.view.sel()) > 1:
```
FichteFoll: You should check for exactly `== 1` because a view can also have no selection.

FichteFoll:
>Should mypanel.show(False) hide the panel? Is this useful?

Doesn't seem useful to me when we also have `.hide`. According to the Zen of Python, there should be only one obvious way.

>Would it be reasonable to default the window argument to sublime.active_window

Just request it explicitly, imo.

>Should we raise an error if you call a method of a panel that's been destroyed?

Yes, because behavior won't be "as specified". There are no doc strings, but I would only describe the useful case and not "does nothing when the panel has been destroyed" since that's irregular behavior.

FichteFoll:
All right, LGTM.

Issue #2, opened 2018-04-03
Settings abstraction


FichteFoll:
An abstractions of the built-in `sublime.Settings` class seems in order, as the following features would be quite useful:

- [x] Registering a callback that is only called when settings of interest *actually* changed, not when the files were reloaded and *could have* changed.
- [x] Providing a default value at instantiation or at a static location (e.g. through `sublime.decode_value(sublime.load_resource(path_to_bundled_settings_file))` rather than having to repeat the default every time the value is accessed.
- [x] Ability to determine *which* setting changed exactly and what the previous value was, if desired.
- [x] Should function independently of whether the underlying settings object is a view or a file settings object.

A related implementation can be found here: https://github.com/SublimeText/InactivePanes/blob/ca417aa6699432ffbbd716ea44e8d431a8d5818f/settings/__init__.py
Downside of that implementation is the complex attribute-to-setting-name mapping that shouldn't need to exist.

Thom1729:
Can't default values be handled by bundling a default `Name.sublime-settings` file with the package?

Thom1729:
As of #15, would you agree that we've handled the known use cases?

FichteFoll:
Agreed. All the points are checked, so this should be finished until we find something else, which would be a new issue.

PR #7, opened 2018-04-05
Added "fancy settings" with bracket access and better subscription.

Thom1729:
For #2.

The name could be changed. I'd be a bit leery of simply calling it “Settings”.

The `subscribe` method takes a “selector”, which maps the current state of a `FancySettings` to some value. (If `selector` is a string, it is interpreted as the name of a top-level setting.) When the `Settings` changes, if and only if the selected value changes, the given callback will be invoked with the new and old value as arguments. This is what I'd need for JS Custom, and I think it should handle most use cases.

Default values are not handled in this PR, though they should be easily compatible with this implementation.

The built-in `Settings` class supplies `None` for a nonexistent key instead of raising a `KeyError`. Should we follow suit or be stricter? Myself, I'd generally lean toward stricter, because the user can always provide a default value in `get`. This PR is loose, like the builtin.

It would be nice to be able to enumerate settings, but I don't think the Sublime API allows for this.

### DIFF
--- a/st3/sublime_lib/fancy_settings.py
+++ b/st3/sublime_lib/fancy_settings.py
@@ -1,0 +1,47 @@
+class FancySettings():
+    def __init__(self, settings):
+        self.settings = settings
+
+    def get(self, name, default=None):
+        return self.settings.get(name, default)
+
+    def set(self, name, value):
+        self.settings.set(name, value)
+
+    def erase(self, name):
+        self.settings.erase(name)
+
+    def has(self, name):
+        return self.settings.has(name)
+
+    def __getitem__(self, key):
+        return self.get(key)
+
+    def __setitem__(self, key, value):
+        self.set(key, value)
+
+    def __delitem__(self, key):
+        self.erase(key)
+
+    def __contains__(self, item):
+        return self.has(item)
+
+    def subscribe(self, key, selector, callback):
+        if isinstance(selector, str):
+            key = selector
+            selector = lambda this: this[key]
+
+        previous_value = selector(self)
+
+        def onchange():
+            nonlocal previous_value
+            new_value = selector(self)
+
+            if new_value != previous_value:
+                callback(new_value, previous_value)
+                previous_value = new_value
+
+        self.settings.add_on_change(key, onchange)
+
+    def unsubscribe(self, key):
+        self.settings.clear_on_change(key)
--- a/st3/sublime_lib/fancy_settings.py
+++ b/st3/sublime_lib/fancy_settings.py
@@ -1,3 +1,5 @@
+from uuid import uuid4
+
 def isiterable(obj):
     try:
         iter(obj)
@@ -40,7 +42,7 @@ class FancySettings():
     def __contains__(self, item):
         return self.has(item)
 
-    def subscribe(self, key, selector, callback):
+    def subscribe(self, selector, callback):
         if callable(selector):
             selector_fn = selector
         elif isinstance(selector, str):
@@ -60,7 +62,9 @@ class FancySettings():
                 callback(new_value, previous_value)
                 previous_value = new_value
 
+        key = str(uuid4())
         self.settings.add_on_change(key, onchange)
+        return key
 
     def unsubscribe(self, key):
         self.settings.clear_on_change(key)
--- a/st3/sublime_lib/fancy_settings.py
+++ b/st3/sublime_lib/fancy_settings.py
@@ -8,17 +8,17 @@ def isiterable(obj):
         return False
 
 class FancySettings():
-    def __init__(self, settings):
+    def __init__(self, settings, defaults={}):
         self.settings = settings
+        self.defaults = defaults
 
     def get(self, name, default=ARGUMENT_NOT_GIVEN):
-        if default is ARGUMENT_NOT_GIVEN:
-            if self.settings.has(name):
-                return self.settings.get(name)
-            else:
-                raise KeyError(name)
+        if self.settings.has(name):
+            return self.settings.get(name)
+        elif default is not ARGUMENT_NOT_GIVEN:
+            return self.defaults.get(name, default)
         else:
-            return self.settings.get(name, default)
+            return self.defaults.get(name)
 
     def set(self, name, value):
         self.settings.set(name, value)
--- a/st3/sublime_lib/fancy_settings.py
+++ b/st3/sublime_lib/fancy_settings.py
@@ -45,19 +45,22 @@ class FancySettings():
         else:
             return default
 
-    def setdefault(key, default=None):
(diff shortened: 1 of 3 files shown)


st3/sublime_lib/fancy_settings.py
```diff
@@ -0,0 +1,47 @@
...
+
+        def onchange():
+            nonlocal previous_value
+            new_value = selector(self)
+
+            if new_value != previous_value:
+                callback(new_value, previous_value)
+                previous_value = new_value
+
+        self.settings.add_on_change(key, onchange)
```
FichteFoll: I prefer using `functools.partial` over closures in situations like this, because closures create a new function for each invocation, whereas partial just binds a parameter. I would define `_on_change` as a private method then.

st3/sublime_lib/fancy_settings.py
```diff
@@ -0,0 +1,47 @@
...
+    def __setitem__(self, key, value):
+        self.set(key, value)
+
+    def __delitem__(self, key):
+        self.erase(key)
+
+    def __contains__(self, item):
+        return self.has(item)
+
+    def subscribe(self, key, selector, callback):
```
FichteFoll: What is the purpose of `key` here? Imho we should try to take the `add_on_change` key away from the user so they don't have to bother with it. Besides, if you fall back the key to `selector` when it's a string, it quickly becomes non-unique.

st3/sublime_lib/fancy_settings.py
```diff
@@ -0,0 +1,47 @@
...
+        if isinstance(selector, str):
```
FichteFoll: I suppose allowing sequences of strings as the selector and building the selector based on `selector = lambda this: {key: this[key] for key in selector}` would also be useful?

FichteFoll:
Raising `KeyError` for keys that the settings object does not have (`settings.has(key) is False`) should probably be added since it's the standard Python behavior. Returning `None` always would be akin to a `defaultdict`.

If defaults are provided to the instance in one way or another, the `.has` call can be skipped anyway.

st3/sublime_lib/fancy_settings.py
Thom1729: In this case, I think we need a closure (and a separate function instance per subscription) because we need to keep track of the last selected value.

st3/sublime_lib/fancy_settings.py
Thom1729: `key` is there solely so that users can unsubscribe. Setting `key` when `selector` was a string was a mistake; I've fixed it.

Thom1729:
I've implemented `KeyErrors` for `get` and `erase`.

st3/sublime_lib/fancy_settings.py
Thom1729: Implemented.

st3/sublime_lib/fancy_settings.py
FichteFoll: True, I missed that it was being modified for some reason. Yeah, I guess we need it then.

Thom1729:
Errors should now be raised in the right places.

I don't think that we can implement any of the `collections.abc` abstract classes because we don't have `__iter__`. I would love to implement that, but I have no idea how. I'm not sure it's possible over the API.

st3/sublime_lib/fancy_settings.py
Thom1729: That is a much better idea.

Thom1729:
`subscribe` no longer takes a `key` argument. Instead, it generates a UUID, uses that as the key, and returns it for future use with `unsubscribe`.

FichteFoll:
I meant the ABC class to check whether an object is iterable using isinstance.

Thom1729:
The behavior used here is [more general](https://stackoverflow.com/questions/1952464/in-python-how-do-i-determine-if-an-object-is-iterable). It may be that we don't need that level of generality, though.

Thom1729:
In order to bring FancySettings closer in line with [the standard dict interface](https://docs.python.org/3/library/stdtypes.html#dict), I've added a few methods (`pop`, `setdefault`, and `update`) and changed some argument names from `name` to `key`. At this point, we're only "missing" the methods that can't be supported by the underlying implementation: `__len__`, `__iter__`, `clear`, `copy`, `fromkeys`, `items`, `keys`, `popitem`, and `values`, plus the comparison operators.

The `set`, `erase`, and `has` methods are not part of the standard interface and provide no additional functionality. Is there any reason to keep them?

On an unrelated note, is there a motivating example somewhere for default values? I sort of understand that they could be useful for view settings and such, but I can't think of a concrete use case.

FichteFoll:
You could also subclass `collections.abc.MutableMapping` and get the default implementations from that.

>The set, erase, and has methods are not part of the standard interface and provide no additional functionality. Is there any reason to keep them?

That's a good point. I say we remove them. This also more clearly separates the underlying API from our abstraction.

>On an unrelated note, is there a motivating example somewhere for default values? I sort of understand that they could be useful for view settings and such, but I can't think of a concrete use case.

Also a good point. On occasions, the default settings files haven't been loaded by the time the settings are accessed (despite the API being ready), so you can't always rely on them. I haven't been able to reliably reproduce this and I also haven't encountered it in a long while, but that could also be because I always include a default value in any `settings.get` call since I've had these annoying issues.
If that wasn't the case, we probably wouldn't need a default mapping as the bundled settings file should already have them.
For user settings, I suppose it is easy enough to specify the default when getting the setting, but the convenience of just accessing an item of the object without having to worry about whether there will always be a value seems quite useful regardless.

FichteFoll:
Hm, I guess `erase` actually behaves slightly different because it doesn't raise for a non-existant setting. Do you think we need it or can users just work around that by checking whether the setting exists or wrap it in a try-catch?

Thom1729:
You can use `fancy_settings.pop(key, None)`.

st3/sublime_lib/fancy_settings.py
```diff
@@ -0,0 +1,87 @@
...
+            return default
+
+    def setdefault(key, default=None):
+        if key in self:
+            return self.get(key)
+        else:
+            self.set(key, default)
+            return default
+
+    def update(*pairs, **kwargs):
```
FichteFoll: This implementation of `update` is not correct.

It accepts either a **single** parameter that is a `dict` or a sequence of key-value pairs, or no positional parameter.
It then applies updates provided through kwargs.

Thom1729: Fixed.

FichteFoll:
LGTM for now. I'll take a deeper look eventually when I find time to test out all of this, but merge away.

PR #12, opened 2018-04-10
Added ViewTestCase -- a subclass of unittest.TestCase.

Thom1729:
`ViewTestCase` has predefined `setUp` and `tearDown` methods and a new `assertContents` method. This should factor out some of the boilerplate from our tests, and I think that this would be a widely useful feature to expose, because most unit tests for Sublime packages would probably involve creating a view, doing something to it, checking the results, and then tearing it down.

The name `assertContents` isn't quite in the style of the other `TestCase` methods. More correct would be `assertViewContentsEqual`, which seems a bit verbose. But maybe that would work best, along with an `assertViewContentsNotEqual`.

### DIFF
--- a/st3/sublime_lib/testing.py
+++ b/st3/sublime_lib/testing.py
@@ -1,0 +1,19 @@
+import sublime
+
+from unittest import TestCase
+
+class ViewTestCase(TestCase):
+    def setUp(self):
+        self.view = sublime.active_window().new_file()
+
+    def tearDown(self):
+        if self.view:
+            self.view.set_scratch(True)
+            self.view.window().focus_view(self.view)
+            self.view.window().run_command("close_file")
+
+    def assertContents(self, text):
+        self.assertEqual(
+            self.view.substr(sublime.Region(0, self.view.size())),
+            text
+        )
--- a/tests/test_fancy_settings.py
+++ b/tests/test_fancy_settings.py
@@ -1,20 +1,14 @@
 import sublime
 from sublime_lib import FancySettings
 
-from unittest import TestCase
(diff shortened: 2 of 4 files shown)


FichteFoll:
Since tests inside ST need to be run with UnitTesting, wouldn't it make sense to add this class there?

st3/sublime_lib/testing.py
```diff
@@ -0,0 +1,19 @@
...
+
+class ViewTestCase(TestCase):
+    def setUp(self):
+        self.view = sublime.active_window().new_file()
+
+    def tearDown(self):
+        if self.view:
+            self.view.set_scratch(True)
+            self.view.window().focus_view(self.view)
+            self.view.window().run_command("close_file")
```
FichteFoll: Isn't `close_file` a view command? If so, you could just run `self.view.run_command("close_file")`.

Thom1729:
> Since tests inside ST need to be run with UnitTesting, wouldn't it make sense to add this class there?

[Yeah, probably.](https://github.com/SublimeText/UnitTesting/issues/89)

FichteFoll:
I'm closing this then. You can commit the change to use the upstream helper directly without a PR.

gerardroche:
Tests don't necessarily have to be run with UnitTesting, it's just a test runner. Sublime Text test cases could be test runner agnostic. The test runners would then choose to enhance those test cases to be package specific, leave them as they are, or provide alternate test runner specific test cases.

But I think it's pretty much academic. <strike>It wouldn't make sense</strike> It would be insane for someone to write another ST test runner package.

PR #13, opened 2018-04-11
ViewStream implements TextIOBase

Thom1729:
- ViewStream now extends [io.TextIOBase](https://docs.python.org/3/library/io.html#io.TextIOBase).
- `_check_selection` will raise if the selection is nonempty, because this could produce surprising behavior with `write` and `tell`.
- `seek` is refactored to avoid redundant checks.
- New `tell` method gives the position of the cursor.
- `read`, `readline`, and `detach` exist and raise `io.UnsupportedOperation`. We could implement `read` and `readline` if we want to.
- Documentation and tests.

### DIFF
--- a/st3/sublime_lib/view_stream.py
+++ b/st3/sublime_lib/view_stream.py
@@ -1,8 +1,9 @@
 from sublime import Region
-from io import SEEK_SET, SEEK_CUR, SEEK_END, TextIOBase, UnsupportedOperation
+from io import SEEK_SET, SEEK_CUR, SEEK_END, TextIOBase
 
-"""A writable text stream encapsulating a `sublime.View` object."""
+
 class ViewStream(TextIOBase):
+    """A writable text stream encapsulating a `sublime.View` object."""
 
     def __init__(self, view):
         self.view = view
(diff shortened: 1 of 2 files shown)


st3/sublime_lib/view_stream.py
```diff
@@ -1,21 +1,32 @@
...
+        elif len(self.view.sel()) > 1:
             raise ValueError("The underlying view has multiple selections.")
+        elif not self.view.sel()[0].empty():
+            raise ValueError("The underlying view's selection is not empty.")
 
     def _check_is_valid(self):
         if not self.view.is_valid():
             raise ValueError("The underlying view is invalid.")
 
+    """Insert the string <var>s</var> into the view and return the number of
```
FichteFoll: docstrings should to go inside the method definition.

I've never seen `<var>s</var>` used before (in Python at least). Does this follow a standard and, if so, which?

FichteFoll:
Good call.

I say we implement `read` and `readline`. They are cheap to implement and may end up being useful.

st3/sublime_lib/view_stream.py
Thom1729: That's HTML. It might not make any sense in a pydoc. Is there a good reference somewhere for pydoc conventions?

PR #14, opened 2018-04-11
Added settings argument to OutputPanel.

Thom1729:
You can pass a dictionary of settings to `__init__` to set them in the panel's view.

In addition to handling syntaxes and such, I'm hoping that this will suffice in place of more specific methods to handle error regexps.

### DIFF
--- a/st3/sublime_lib/output_panel.py
+++ b/st3/sublime_lib/output_panel.py
@@ -2,11 +2,16 @@ from .view_stream import ViewStream
 
 
 class OutputPanel(ViewStream):
-    def __init__(self, window, name):
+    def __init__(self, window, name, *, settings=None):
         super().__init__(window.get_output_panel(name))
 
         self.window = window
         self.name = name
+
+        if settings:
+            view_settings = self.view.settings()
+            for key, value in settings.items():
+                view_settings.set(key, value)
 
     @property
     def full_name(self):
--- a/st3/sublime_lib/output_panel.py
+++ b/st3/sublime_lib/output_panel.py
@@ -2,16 +2,22 @@ from .view_stream import ViewStream
 
 
 class OutputPanel(ViewStream):
-    def __init__(self, window, name, *, settings=None):
+    def __init__(self, window, name, *,
+        settings=None,
+        read_only=None
+    ):
         super().__init__(window.get_output_panel(name))
 
         self.window = window
         self.name = name
 
-        if settings:
+        if settings is not None:
             view_settings = self.view.settings()
             for key, value in settings.items():
                 view_settings.set(key, value)
+
+        if read_only is not None:
+            view.set_read_only(read_only)
 
     @property
     def full_name(self):
--- a/st3/sublime_lib/output_panel.py
+++ b/st3/sublime_lib/output_panel.py
@@ -2,7 +2,8 @@ from .view_stream import ViewStream
 
 
 class OutputPanel(ViewStream):
-    def __init__(self, window, name, *,
+    def __init__(
+        self, window, name, *,
         settings=None,
         read_only=None
     ):
@@ -17,7 +18,7 @@ class OutputPanel(ViewStream):
                 view_settings.set(key, value)
 
         if read_only is not None:
-            view.set_read_only(read_only)
+            self.view.set_read_only(read_only)
 
     @property
     def full_name(self):
--- a/st3/sublime_lib/output_panel.py
+++ b/st3/sublime_lib/output_panel.py
@@ -4,10 +4,14 @@ from .view_stream import ViewStream
 class OutputPanel(ViewStream):
     def __init__(
         self, window, name, *,
+        force_writes=False,
         settings=None,
         read_only=None
     ):
-        super().__init__(window.get_output_panel(name))
+        super().__init__(
+            window.get_output_panel(name),
+            force_writes=force_writes
+        )
 
         self.window = window
         self.name = name
--- a/st3/sublime_lib/view_stream.py
+++ b/st3/sublime_lib/view_stream.py
@@ -16,8 +16,13 @@ class ViewStream():
         if not self.view.is_valid():
             raise ValueError("The underlying view is invalid.")
 
+    def _check_writable(self):
+        if self.view.is_read_only():
+            raise ValueError("The underlying view is read-only.")
+
     def write(self, s):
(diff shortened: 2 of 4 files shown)


FichteFoll:
I wonder if it wouldn't be better to set the syntax with set_syntax_file. Iirc that has a few advantages.

Thom1729:
Do you recall what advantages?

Thom1729:
I've added a `read_only` kwarg as well. (The `read_only` property is completely orthogonal to settings.)

On a related note, it would probably be useful to have a handy view-factory function somewhere. Something like:

    def create_view(window, *,
        settings=None,
        name=None,
        read_only=None,
        scratch=None
    ):
        view = window.new_file()

        ...
```

I'm not quite sure where this would go in the package.

Finally, on the topic of `read_only`, it occurs to me that `ViewStream` doesn't check for this. The simplest/safest thing to do would be to raise `ValueError` (or `IOError`?) if you try to `write` a `read_only` view.

File: st3/sublime_lib/output_panel.py [hunk 1]
```diff
@@ -2,12 +2,23 @@
 
 
 class OutputPanel(ViewStream):
-    def __init__(self, window, name):
+    def __init__(self, window, name, *,
+        settings=None,
```
FichteFoll: Please use visual indentation when there are arguments on the first line.

Don't we run flake8 on this? It should complain here.

st3/sublime_lib/output_panel.py
```diff
@@ -2,12 +2,23 @@
...
         self.window = window
         self.name = name
 
+        if settings is not None:
+            view_settings = self.view.settings()
+            for key, value in settings.items():
+                view_settings.set(key, value)
+
+        if read_only is not None:
+            view.set_read_only(read_only)
```
FichteFoll: In my implementation, I used to unset read-only for the methods that operate on the contents. I believe this is in order here as well, or do you expect the user to handle that manually? In that case I'd rather have them manage read-only entirely by themselves from outside of the class's implementation.

FichteFoll:
>On a related note, it would probably be useful to have a handy view-factory function somewhere.

How do we feel about an augmented `View` class? I think we discussed it before, but am not sure what we came onto. If this was implemented, that should be one of its constructors (probably as a classmethod) so that the main constructor can still be used to wrap actual views.
For now, we can however just dump this in a module like `sublime_lib.view` or `sublime_lib.view_utils` or something.

FichteFoll:
>Finally, on the topic of read_only, it occurs to me that ViewStream doesn't check for this

I touch on this in my review, as I read the changes before your comment. First though was introducing a `ProtectedViewStream` subclass that handles unwrapping read-only for text operations, but that would end up messing with `OutputPanel`. I do believe that having this functionality outside of `OutputPanels` would be useful, however, so we should either add a `read_only` flag to `ViewStream` natively or consider a separate class (in which case the `OutputPanel` implementation becomes wonky).

st3/sublime_lib/output_panel.py
Thom1729: I haven't automated it yet on my end.

Thom1729:
> I touch on this in my review, as I read the changes before your comment. First though was introducing a ProtectedViewStream subclass that handles unwrapping read-only for text operations, but that would end up messing with OutputPanel. I do believe that having this functionality outside of OutputPanels would be useful, however, so we should either add a read_only flag to ViewStream natively or consider a separate class (in which case the OutputPanel implementation becomes wonky).

I think that there are two possible sane behaviors. When you `write` to a read-only view, it should either raise an error or unset, write, set. (That is, silently failing to write is not desirable.) We could handle this pretty easily as a flag in the `ViewStream` constructor (`force_writes` or somesuch). The OutputPanel constructor could take the same flag and pass it along, thus maintaining the same behavior. I think that this would be preferable to creating extra classes.

Thom1729:
Example:

```python
def write(self, s):
        """Insert the string <var>s</var> into the view and return the number of
        characters inserted. The string will be inserted immediately before the
        cursor.
        """

        self._check_is_valid()
        self._check_selection()

        if self.view.is_read_only():
            if self.force_writes:
                try:
                    self.view.set_read_only(False)
                    return self._write(s)
                finally:
                    self.view.set_read_only(True)
            else:
                raise ValueError("The underlying view is read-only.")
        else:
            return self._write(s)

    def _write(self, s):
        self.view.run_command('insert', {'characters': s})
        return len(s)
```

FichteFoll:
Sounds good. The `OutputPanel` would then just have a `read_only` flag and if that is set, it would also specify `force_writes` to its super constructor because it doesn't make any sense otherwise.

Thom1729:
> How do we feel about an augmented View class? I think we discussed it before, but am not sure what we came onto. If this was implemented, that should be one of its constructors (probably as a classmethod) so that the main constructor can still be used to wrap actual views.
For now, we can however just dump this in a module like sublime_lib.view or sublime_lib.view_utils or something.

I like the idea of a factory function, but I'm leery of wrapping the `View` class in its entirety. The `ViewStream` class solves a common but specific problem, making a certain set of assumptions. The `ViewBuffer` class is designed to solve a different problem, and its approach is incompatible with `ViewStream`'s. An augmented `View` class couldn't do this; all we could do is add utility functions, which we can do anyway without augmenting `View` itself.

For `FancySettings`, we did basically provide a total replacement for `Settings`, but that just naturally happened as a result of providing a `dict`-like interface. In providing a general-purpose augmented `View`, we'd end up wrapping a giant pile of methods without improving upon most of them.

FichteFoll:
You can leave the augmented view out for now. I have something in mind that you'll likely understand easier when you see it.

---

Regarding the PR at hand, you need to fix `.clear` also. After that I think we are ready to merge.

FichteFoll:
```
[02:44] jps: if you just set the syntax setting, as opposed to using assign_syntax, then the syntax won't be applied properly
[02:45] jps: syntax highlighting will be fine, but the preferences won't get setup for the view
[02:45] jps: e.g., if you want the settings  in C++.sublime-settings to be applied, then you need to use assign_syntax()
```

Thus, I believe we need that as a separate argument.

Thom1729:
I have some thoughts on the syntax thing. We'll definitely want it as a separate argument, but I'll just merge this for now.

I will be traveling through Monday.

PR #15, opened 2018-04-12
FancySettings can be constructed with a name, and later saved.

Thom1729:
This will probably be the typical use case: `settings = FancySettings("MyPackage.sublime-settings")`.

### DIFF
--- a/st3/sublime_lib/fancy_settings.py
+++ b/st3/sublime_lib/fancy_settings.py
@@ -1,3 +1,4 @@
+import sublime
 from uuid import uuid4
 from collections.abc import Mapping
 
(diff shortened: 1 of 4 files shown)


st3/sublime_lib/fancy_settings.py
```diff
@@ -70,6 +79,12 @@ def update(self, other=[], **kwargs):
         for key, value in kwargs.items():
             self[key] = value
 
+    def save(self):
+        if self.name is None:
+            raise ValueError("Only named settings can be saved.")
```
FichteFoll: Maybe a `RuntimeError` instead?

Thom1729: I'm a bit confused as to the precise distinction between `RuntimeError` and `ValueError`. There's no standard exception for "this object is in the wrong state for you to call this method".

FichteFoll: The description for `ValueError` is:

>Raised when a built-in operation or function receives an argument that has the right type but an inappropriate value, and the situation is not described by a more precise exception such as IndexError.

If there is no particular exception fitting your case, you usually go up the hierarchy until you arrive at `RuntimeError`. An `UnexpectedStateException` doesn't exist, so that seems like the best fit.

Thom1729: Actually, in hindsight I think we should just split the name-based settings into a subclass:

```python
class NamedFancySettings():
    def __init__(self, name, defaults={}):
        super().__init__(sublime.load_settings(name), defaults)
        self.name = name

    def save(self):
        sublime.save_settings(self.name)
```

FichteFoll: Yeah, I like that.

PR #18, opened 2018-04-14
Syntax utilities improvements

FichteFoll:
Add some rudimentary YAML value parsing and a couple more tests.

### DIFF
--- a/st3/sublime_lib/syntax.py
+++ b/st3/sublime_lib/syntax.py
@@ -14,13 +14,16 @@ def _parse_yaml_value(value):
     if value.startswith("'"):
         return value[1:-1].replace("''", "'")
     elif value.startswith('"'):
-        # This is a simplifcation but one that should do for now
-        return re.sub(r"\\(.)", r"\1", value[1:-1])
+        # JSON and YAML quotation rules are very similar, if not identical
+        return sublime.decode_value(value)
     elif value == "true":
         return True
     elif value == "false":
         return False
+    elif value == "null":
+        return None
     else:
+        # Does not handle numbers because we don't expect any
         return value
 
 
--- a/st3/sublime_lib/syntax.py
+++ b/st3/sublime_lib/syntax.py
@@ -1,22 +1,27 @@
 import sublime
 
 import re
+from collections import namedtuple
 
 __all__ = ['list_syntaxes', 'get_syntax_for_scope']
 
 
-class SyntaxInfo():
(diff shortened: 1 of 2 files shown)


st3/sublime_lib/syntax.py
```diff
@@ -1,38 +1,52 @@
...
+
+SyntaxInfo = namedtuple('SyntaxInfo', ['path', 'name', 'scope', 'hidden'])
+SyntaxInfo.__new__.__defaults__ = (None, None, False)  # type: ignore
+
+
+def _parse_yaml_value(value):
+    if value.startswith("'"):
+        return value[1:-1].replace("''", "'")
+    elif value.startswith('"'):
+        # This is a simplifcation but one that should do for now
```
Thom1729: Can we just use `sublime.decode_value` here? I'd have to compare the JSON and YAML specs to be sure, but I think this should work.

st3/sublime_lib/syntax.py
```diff
@@ -1,38 +1,52 @@
...
+        return re.sub(r"\\(.)", r"\1", value[1:-1])
+    elif value == "true":
+        return True
+    elif value == "false":
+        return False
+    else:
```
Thom1729: We should handle `null` as well.

st3/sublime_lib/syntax.py
```diff
@@ -1,38 +1,52 @@
...
-    __slots__ = [ 'path', 'name', 'scope', 'hidden' ]
-    
-    def __init__(self, path, name=None, scope=None, hidden=False):
-        self.path = path
-        self.name = name
-        self.scope = scope
-        self.hidden = hidden
+
+SyntaxInfo = namedtuple('SyntaxInfo', ['path', 'name', 'scope', 'hidden'])
+SyntaxInfo.__new__.__defaults__ = (None, None, False)  # type: ignore
```
Thom1729: This looks good; I actually started with `namedtuple` and stopped using it because I didn't know about `__defaults__`. Just to be sure, this is for sure a stable feature?

FichteFoll: Yes, it's referenced on https://docs.python.org/3/reference/datamodel.html.

st3/sublime_lib/syntax.py
FichteFoll: Good idea. Yes, that should work better as it supports unicode escapes.

PR #19, opened 2018-04-15
Use decorators to ensure valid ViewStream state

FichteFoll:
Had this idea while reviewing whatever PR it was that added the checks. Initially I had each check as a generator, but then I realized I'd end up wrapping a method up to 3 times and that didn't seem right, so I rewrote the guards into normal checks (that raise exceptions) and added a "meta" guard that these checks are passed to. Also added context manager support to unset the read-only flag.

I guess this PR is quite controversial, so let me know what you think.

Advantages:
- Communicates more clearly what state invariants are required for a method to run successfully.
- Gets rid of internal functions that were wrapped with `_wrap_read_only`.

Disadvantages:
- Higher complexity and less obvious for developers not used to decorators.

### DIFF
--- a/st3/sublime_lib/output_panel.py
+++ b/st3/sublime_lib/output_panel.py
@@ -1,4 +1,5 @@
 from .view_stream import ViewStream
+from .view_stream import _guard, _guard_validity
 
 
 class OutputPanel(ViewStream):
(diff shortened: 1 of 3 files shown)


Thom1729:
I love the idea. I think that decorators are the perfect way to express preconditions.

I do think that it might look clearer “nested”:

```python
@guard_validity
@guard_selection
def read(self, size):
```

There would be more levels of nesting; but then, the logic of `_guard` is nontrivial. I suspect that the performance difference would be negligible (in whichever direction), especially given that most of these methods are making API calls anyway.

FichteFoll:
Yeah, I guess you are right. The number of wrappers created is manageable and the functions need to be called anyway, even with the current guard helper.

FichteFoll:
Good changes. :+1: I forgot you can your methods/decorators within a class's definition after you defined them, which makes this very elegant.

PR #27, opened 2018-04-23
Fixed bugs in ViewStream.write.

Thom1729:
The size returned from `write` is now measured, because it can be different than the length of the string argument.

In addition, the inserted text will not be auto-indented. This could conceivably be an option (like `force_write`).

### DIFF
--- a/st3/sublime_lib/view_stream.py
+++ b/st3/sublime_lib/view_stream.py
@@ -100,8 +100,13 @@ class ViewStream(TextIOBase):
         characters inserted. The string will be inserted immediately before the
         cursor.
         """
-        self.view.run_command('insert', {'characters': s})
-        return len(s)
+        # This is a hack to get around auto-indentation.
+        old_size = self.view.size()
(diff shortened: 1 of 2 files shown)


FichteFoll:
>because it can be different than the length of the string argument.

When does that happen?

Thom1729:
> When does that happen?

Tab conversion.

> Don't think we ever want to unset this, do we?

I don't see any immediate need. The option is open if we need it in the future.

FichteFoll:
>>When does that happen?

>Tab conversion.

Ugh. Didn't think about this. This seems problematic from an API PoV because you don't really expect a stream's `write` method to end up writing *more* characters than you handed it. I guess it's still better to return the actual number of written chars though, e.g. for when you want to seek backwards.

Please add this to the docstring or at least a comment (for now) as it can be unexpected for users and we shouldn't forget about it either.

>The option is open if we need it in the future.

Well, let's just forget about it for now, then.

Thom1729:
Should be good now.

PR #28, opened 2018-04-26
Added new features to OutputPanel.

Thom1729:
Add new methods `is_visible` and `toggle_visibility`. These would definitely be used by, e.g., SublimeLinter.

Use `create_output_panel` instead of `get_output_panel`. The latter is apparently deprecated. The former supports an `unlisted` parameter, and now so do we.

We should probably explicitly address what happens when the user creates a panel that already exists (i.e. whose name is the name of an existing panel). We could:

- Explicitly destroy the old panel, then create a new one.
- Re-use the existing view, but apply the given options.
- Raise an error.
- Any of the above depending on an argument.

I'm not quite sure what Sublime does in the present implementation -- specifically how it may differ from simply re-using the view. I do know that the `unlisted` argument is ignored if a panel by the given name already exists.

### DIFF
--- a/st3/sublime_lib/output_panel.py
+++ b/st3/sublime_lib/output_panel.py
@@ -6,12 +6,13 @@ class OutputPanel(ViewStream):
     def __init__(
         self, window, name, *,
         force_writes=False,
+        unlisted=False,
         **kwargs
     ):
         validate_view_options(kwargs)
 
         super().__init__(
-            window.get_output_panel(name),
+            window.create_output_panel(name, unlisted),
             force_writes=force_writes
         )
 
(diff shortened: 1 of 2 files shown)


FichteFoll:
Always raising an error would certainly be wrong, as it prevents you from re-using the same panel name at all, unless you destroyed it before.

Re-using the existing but applying options is weird too, because it may interfere with whatever was set previously, in the case the panel is "created" concurrently. No idea whether this has use cases as you're likely to run into races without locking panel access.

Iirc, the panel's contents are cleared once you "create" it anew. I suggest we add the functionality to take over an existing panel with all its contents and existing options (and settings) through an `exist_ok` parameter. Here the user specifically states that he is okay with re-using the existing panel and wants to inherit its setting (in whatever state they are).

When `exist_ok` is false, just create a new one imo.

Thom1729:
> Re-using the existing but applying options is weird too...

Completely agreed.

It sounds like there may be two distinct use cases. In one case, the user wants to attach to an existing output panel, whatever state it may be in. In the other, they want to create an exclusive new panel with a range of initialization options. This is a lot of difference to hang on one flag argument. Maybe we should handle these two cases with two different methods:

- `OutputPanel(window, name)` creates a new object attached to an existing view.
- `OutputPanel.create(window, name, unlisted, *more, **args)` would destroy any existing panel, create a new one, set a bunch of options, and return an `OutputPanel` object attached to the new view.

One arguable benefit of this is that `OutputView()` now resembles the parent's contructor, `ViewStream`, which works with an existing given view; whereas `OutputPanel.create` now resembles the `view_utils.new_view` function. There's a symmetry there that makes a lot of sense.

There might still be a use case for finding a panel if it exists, and otherwise creating a new one with arguments. This is a more complicated and possibly confusing operation, but it could be implemented as `OutputView.find_or_create` or some other name that explicitly highlights the conditional behavior.

PR #30, opened 2018-04-30
Implemented default settings logic in new DefaultFancySettings class.

Thom1729:
This implementation is inspired by `defaultdict`, and I think it might be more consistent with existing practice.

When you index a `FancySettings` with a nonexistent key, it will call `self.__missing__(key)`  and return the result. The base implementation simply raises KeyError. A new `DefaultFancySettings` subclass takes a `defaults` parameter in its constructor and overrides `__missing__` to set and return a default value. This is basically how `defaultdict` works (although that class takes a `default_factory` function, not a `defaults` dict).

The major difference is that `NamedFancySettings` doesn't accept manual defaults, but the preferred way to supply defaults to a named settings object is to provide a default settings file. View-specific settings and other "anonymous" settings objects can use `DefaultFancySettings` instead.

### DIFF
--- a/st3/sublime_lib/fancy_settings.py
+++ b/st3/sublime_lib/fancy_settings.py
@@ -172,7 +172,7 @@ class NamedFancySettings(FancySettings):
 
 
 class DefaultFancySettings(FancySettings):
-    def __init__(self, settings, defaults={}):
+    def __init__(self, settings, defaults):
         """
         Return a new FancySettings wrapping a given Settings object *settings*.
         """
@@ -180,4 +180,4 @@ class DefaultFancySettings(FancySettings):
         self.defaults = defaults
 
     def __missing__(self, key):
-        return self.defaults[key]
+        return self.setdefault(key, self.defaults[key])
--- a/st3/sublime_lib/fancy_settings.py
+++ b/st3/sublime_lib/fancy_settings.py
@@ -45,7 +45,18 @@ class FancySettings():
         self.settings = settings
 
     def __getitem__(self, key):
-        """Return the setting named *key*. Raises a KeyError if there is no such setting."""
+        """
+        Return the setting named *key*. Raises a KeyError if there is no such
+        setting.
+
+        If a subclass of FancySettings defines a method `__missing__()` and
+        `key` is not present, the `d[key]` operation calls that method with the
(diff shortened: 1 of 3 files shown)


st3/sublime_lib/fancy_settings.py
```diff
@@ -158,12 +160,24 @@ class NamedFancySettings(FancySettings):
...
+class DefaultFancySettings(FancySettings):
+    def __init__(self, settings, defaults):
+        """
+        Return a new FancySettings wrapping a given Settings object *settings*.
+        """
+        super().__init__(settings)
+        self.defaults = defaults
+
+    def __missing__(self, key):
+        return self.setdefault(key, self.defaults[key])
```
FichteFoll: Do we actually need `setdefault`? What's the motivation behind it over `get`?

For defaultdicts it makes sense since the default factory function would, say, create a new list so that you can do stuff like 
```py
ddict = defaultdict(list)
ddict['nonexistant'] += [1]
ddict['nonexistant'] += [2]
ddict['nonexistant'] == [1, 2]
```
, but we don't need this here since we return existing references.

Either way, if `setdefault` was to be used, its side-effect should be documented.

Thom1729: I've modified `__missing__` to work exactly the same as in `defaultdict`. When `__missing__` is called, it will unconditionally set the value to the default (even if it had another value). This shouldn't matter much because users probably shouldn't manually call `__missing__`, but at least it's consistent.

Using `get` would make sense as well, but all else being equal it's probably best to stick with the existing convention.

Documentation has been updated.

st3/sublime_lib/fancy_settings.py
```diff
@@ -172,12 +183,36 @@ def save(self):
...
+
+    This class generally should not be used with named settings files. Instead,
+    package developers should provide settings files populated with defaults.
+    """
+
     def __init__(self, settings, defaults):
         """
-        Return a new FancySettings wrapping a given Settings object *settings*.
+        Return a new DefaultFancySettings wrapping a given Settings object
+        *settings*, with a given dict-like object *defaults* of default values.
```
FichteFoll: The proper term would be "mapping", but dict-like works.

PR #32, opened 2018-05-11
Sort listed syntaxes by load order. Return last matching syntax for s…

Thom1729:
…cope.


FichteFoll:
I'd probably have used a tuple as the sort key, but this works as well. :+1:

PR #33, opened 2018-05-31
Renamed FancySettings to SettingsDict.

Thom1729:
Since we ended up basically implementing the `dict` interface, `SettingsDict` might be a more descriptive name than `FancySettings`. Looking for feedback.


FichteFoll:
Definitely agreed.

PR #34, opened 2018-06-13
[WIP] Add Sphinx documentation

Thom1729:
This PR is not ready, but feedback on the approach is needed.

I've set up Sphinx to autogenerate API documentation. Questions and concerns:

- I'm not entirely satisfied with the theme. Something less heavyweight might be better.
- The `source/*.rst` files are mostly autogenerated, and it would be nice not to have to commit them.
- Likewise, some of the binary files (e.g. doctrees) probably don't need to be there.
- The `flags` docs are sketchy. Sphinx's enum autodocs apparently conflict with its module mocking in a strange way.
- It would be nice to auto-link Sublime's API docs.


FichteFoll:
>It would be nice to auto-link Sublime's API docs.

What do you mean with this? The official API docs aren't hotlinkable, unfortunately, except for the classes.

I'll look at this at a later time.

Thom1729:
>  I can set up ci intergration with github pages some time later to have them auto-update

I'm very new to github pages. To clarify, is it possible to configure CI to create the `.html` files and put them somewhere other than `master` where github pages will display them? If so, this definitely seems better.

FichteFoll:
Yes, it'd be pushed to a `gh-pages` branch entirely separate from `master` and we'd get a website for free as well.

FichteFoll:
Here's the script I used last time: https://github.com/Irrational-Encoding-Wizardry/guide.encode.moe/blob/master/.travis/deploy.sh

FichteFoll:
I'm closing this in favor of #35 as this has way too many redundant files. I squashed the commits from this branch and removed most of those.

PR #35, opened 2018-06-15
Docs slim

FichteFoll:
As slimmer version of #34 with generated files (`docs/`) ignored and auto-deployment of build artifacts to https://sublime_lib.github.io/ (once merged into `master`).

(Which hopefully Just Works™.)


Thom1729:
The `.autodoc` file is vestigial; I shouldn't have added it in the first place. (It's related to a custom documentation generator I was working on.)

The `source` directory should probably be renamed (to `docs-source` or something).

The `<module>.rst` files are autogenerated and arguably shouldn't be committed.

Will take another look after lunch.

FichteFoll:
Agree to all of this. I'd actually prefer `docs/source`, next to `docs/html`.

I haven't used autodoc so far, but it seems like you configure it in `config.py`.
I take it then, `index.rst` is the only file we need to edit manually then? There should definitely be some file explaining the scope and motivation or similar that goes beyond simple API docs.

If yes, I'll go ahead and remove the redundant files, rename the folder (and references) and squash the branch so we get a cleaner history. No need to add and remove files we never need.

Thom1729:
Sounds good.

FichteFoll:
Should be good now. Please merge if you don't see issues as I'd like to see whether CI works.

PR #36, opened 2018-06-15
Use undocumented View.close method in `close_view`.

Thom1729:
Much better than the old approach.

### DIFF
--- a/st3/sublime_lib/view_utils.py
+++ b/st3/sublime_lib/view_utils.py
@@ -65,8 +65,7 @@ def close_view(view, *, force=False):
         else:
             raise ValueError('The view has unsaved changes.')
 
-    view.window().focus_view(view)
-    view.window().run_command("close_file")
+    view.close()
 
 
 def validate_view_options(options):
--- a/st3/sublime_lib/view_utils.py
+++ b/st3/sublime_lib/view_utils.py
@@ -50,22 +50,17 @@ def close_view(view, *, force=False):
 
     If the view is invalid (e.g. already closed), do nothing.
 
-    :raise ValueError: if the view has no associated window (e.g. a panel).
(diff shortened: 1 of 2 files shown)


FichteFoll:
Can you confirm that this works as expected? I seem to remember issues with `view.close`, but not exactly which. Maybe when the view is inactive?

Thom1729:
It seems like the behavior of `view.close()` is as follows:

- If the view is invalid or cannot be closed (e.g. a panel), return False.
- If the view has unsaved changes and is not a scratch view, display the "save changes?" dialog. If cancelled, return False.
- Finally, close the view and return True.

In particular, the method can "hang" while the dialog is open. Eventually, it will return True or False to indicate whether the view was closed as a result of the call.

I have tested inactive views and had no problems so far.

The `close_view` method never triggers the dialog, because if there are unsaved changes it will either discard them or raise `ValueError`.

FichteFoll:
Okay, this sounds good. 

I believe we can use the return value of `View.close`, however. We could either raise an exception if it's `False` or just forward its return value. The problem with forwarding would be that we already raise an exception if closing failed due to the view being dirty, so we'd have two ways of signaling failure.

Thom1729:
Raising an exception makes sense. Can you think of any way to test this?

FichteFoll:
Just use direct ST api to make the close call return False?

>- If the view is invalid or cannot be closed (e.g. a panel), return False.

Thom1729:
I realized that most of the error handling in that function was redundant, so now the existing tests will be using the new error path. I also added a test for closing an already-closed view.

FichteFoll:
All right, this is good.

However, unfortunately it seems that the branch got messed up. You can merge once you rebased (as I'll be leaving now).

PR #38, opened 2018-06-16
Remove DefaultSettingsDict

Thom1729:
`collections.ChainMap` can be used instead of `DefaultSettingsDict`, so the latter is unnecessary.

`SettingsDict` now implements `__iter__`, raising [`NotImplementedError`](https://docs.python.org/3/library/exceptions.html#NotImplementedError). Formerly, calling `iter` on a `SettingsDict` would indirectly raise `TypeError`, which was confusing.

In principle, `NotImplementedError` is supposed to be used for methods that are supposed to be implemented in the future. The degree to which this describes `SettingDict.__iter__` depends on one's optimism that the Sublime API will someday permit this. Either way, I think that `NotImplementedError` is the clearest way to indicate that, unusually for a collection, `SettingsDict` is not iterable.

I've added tests checking the `NotImplementedError` and ensuring that `SettingsDict` works in a `ChainMap`.

### DIFF
--- a/st3/sublime_lib/__init__.py
+++ b/st3/sublime_lib/__init__.py
@@ -1,6 +1,5 @@
 from .settings_dict import SettingsDict  # noqa: F401
 from .settings_dict import NamedSettingsDict  # noqa: F401
-from .settings_dict import DefaultSettingsDict  # noqa: F401
 from .view_buffer import ViewBuffer  # noqa: F401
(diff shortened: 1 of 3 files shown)


FichteFoll:
Could you also add a mention for ChainMap in SettingsDict's docstring?

Thom1729:
Good idea.

PR #39, opened 2018-06-18
Documented flags.

Thom1729:
The Sphinx mock module class interacts weirdly with enums, so I wrote a custom mock that provides arbitrary integers for properties that look like constants. With this done, I added a decorator that automatically adds a enum's members to its docstring.


FichteFoll:
:+1:

PR #41, opened 2018-07-11
Removed top-level exports.

Thom1729:
Given the number of items provided by the package, it probably doesn't make sense to export all of them from `__init__`. If there is a sensible strategy for exporting a subset at top level, we can do that instead. For now, this PR removes all top-level exports in preparation for the stable release.


FichteFoll:
I don't consider 5 items a lot.

The problem I have without top-level exports is that in the current structure we use submodules more like Java where we just use different files to separate the implementation, rather than grouping them into relevant sub-modules, with the only exception being `NamedSettingsDict`.

When your modules have the same name as the main thing you're exporting, it becomes an implementation detail and I'd rather not expose that.
```py
from .settings_dict import SettingsDict  # noqa: F401
from .settings_dict import NamedSettingsDict  # noqa: F401
from .view_buffer import ViewBuffer  # noqa: F401
from .view_stream import ViewStream  # noqa: F401
from .output_panel import OutputPanel  # noqa: F401
```

Thom1729:
Top-level exports per se don't bother me; I merely feel that we don't have a coherent policy for them. I would like to either come up with one or punt on the issue so that we can put out a stable release.

Do you think that it would be reasonable to export all public items at top level?

FichteFoll:
>Do you think that it would be reasonable to export all public items at top level?

Can you elaborate on that? I don't think I understand what "public items" mean here.

My policy would go from a different direction and be to hide implementation segmentation on the inside. The `output_panel` module for example only contains the OutputPanel class. From an API point, you don't want to type `from sublime_lib.output_panel import OutputPanel` as that's repetitive and redundant. Thus, it should be elevated by one level, which would in this case be the package root.
On the other hand, the `view_utils` module is a collection of – what the name implies – utilities (functions) with views. There is nothing wrong with `from sublime_lib.view_utils import new_view`. The module is used to group related functions.

PR #42, opened 2018-07-11
Remove ViewBuffer for now.

Thom1729:
Because the useful potential features — buffering of edit operations and cached scopes — haven't been implemented yet. We can re-add it later if appropriate (with tests and docs).


FichteFoll:
I forgot what this was even supposed to be used for.

PR #43, opened 2018-07-11
Add coveragerc.

Thom1729:
Provide more accurate coverage information.


FichteFoll:
I personally find coverage information on test files quite interesting as it allows to see whether tests aren't properly executed, for any reason.

PR #44, opened 2018-07-15
 Construct NamedSettingsDict without explicit .sublime-settings extension.

Thom1729:
Before:

```python
settings = NamedSettingsDict("Preferences.sublime-settings")
settings.name # "Preferences.sublime-settings"
```

After:

```python
settings = NamedSettingsDict("Preferences")
settings.name # "Preferences"
settings.file_name # "Preferences.sublime-settings"
```

### DIFF
--- a/st3/sublime_lib/settings_dict.py
+++ b/st3/sublime_lib/settings_dict.py
@@ -202,7 +202,11 @@ class NamedSettingsDict(SettingsDict):
     def __init__(self, name):
         """Return a new NamedSettingsDict corresponding to the given name."""
 
-        self.name = name
+        if name.endswith('.sublime-settings'):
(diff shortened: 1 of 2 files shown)


FichteFoll:
I'm hesitant about this because of loss of symmetrie with the default implementation (of `sublime.load_settings). Of course, ideally you wouldn't want to specify the extension, but we should probably add compatibility for the case where the extension is provided. I'm sure it would lead to less confusion for users.

Thom1729:
If the user passes the extension, would you prefer to:

- Strip the extension and continue.
- Raise ValueError.

FichteFoll:
I would prefer to have the extension skipped, yes.

PR #46, opened 2018-07-31
Improve readme.

Thom1729:


README.md
```diff
@@ -1,3 +1,27 @@
 # sublime_lib
 
-Utility library for frequently used functionality in Sublime Text library and convenience functions or classes
+An unofficial "standard library" for Sublime Text. Provides a variety of useful convenience features for other packages.
```
FichteFoll: I don't think we need to declare it as "unofficial" when there is no official competition. Not sure if "standard library" is fitting either.

README.md
```diff
@@ -1,3 +1,27 @@
...
+
+Then, anywhere in your package, you can import sublime_lib by name:
+
+```python
+import sublime_lib
+```
+
+## Features
+
+For complete documentation of all features, see the [API documentation](https://sublimetext.github.io/sublime_lib/).
```
FichteFoll: I highly suggest adding an opinionated short list of key features that the library provides. Having to go to yet another page only to find out what this thing actually *does* is not good potential user engagement.

README.md
```diff
@@ -1,3 +1,27 @@
 # sublime_lib
 
-Utility library for frequently used functionality in Sublime Text library and convenience functions or classes
+An unofficial "standard library" for Sublime Text. Provides a variety of useful convenience features for other packages.
+
+## Installation
+
+To make use of sublime_lib in your own package, first declare it as a [dependency](https://packagecontrol.io/docs/dependencies) of your package. Create a file named `dependencies.json` in the root of your package with the following contents:
```
FichteFoll: Might need to mention running the Satisfy Dependencies command.

README.md
Thom1729: Agreed.

PR #47, opened 2018-11-05
Return unsubscribe function from SettingsDict.subscribe

Thom1729:
Currently, `SettingsDict.subscribe` returns the autogenerated subscription key. This return value is undocumented. The also-undocumented `SettingsDict.unsubscribe` method simply wraps `sublime.Settings.clear_on_change`.

With this PR, `SettingsDict.subscribe` returns a function that clears the subscription. This way, we don't have to expose the meaningless autogenerated key or an additional `unsubscribe` method.

### DIFF
--- a/st3/sublime_lib/settings_dict.py
+++ b/st3/sublime_lib/settings_dict.py
@@ -2,6 +2,7 @@ import sublime
 
 from uuid import uuid4
 from collections.abc import Mapping
+from functools import partial
 
 from .collection_utils import projection
 
@@ -182,7 +183,7 @@ class SettingsDict():
 
         key = str(uuid4())
         self.settings.add_on_change(key, onchange)
-        return lambda: self.settings.clear_on_change(key)
+        return partial(self.settings.clear_on_change, key)
 
 
 class NamedSettingsDict(SettingsDict):
--- a/st3/sublime_lib/settings_dict.py
+++ b/st3/sublime_lib/settings_dict.py
@@ -145,8 +145,9 @@ class SettingsDict():
 
     def subscribe(self, selector, callback, default_value=None):
         """
-        Register a calback to be invoked when the a value derived from the
-        settings object changes.
+        Register a callback to be invoked when the a value derived from the
+        settings object changes. Returns a function that when invoked will
+        unregister the callback.
 
         The derived value depends on the type of `selector`:
 
@@ -181,10 +182,7 @@ class SettingsDict():
 
         key = str(uuid4())
         self.settings.add_on_change(key, onchange)
-        return key
-
-    def unsubscribe(self, key):
-        self.settings.clear_on_change(key)
+        return lambda: self.settings.clear_on_change(key)
 
 
 class NamedSettingsDict(SettingsDict):
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,7 @@ SOURCEDIR     = docs/source
 BUILDDIR      = docs
 MODULEDIR     = st3
 
-.PHONY: Makefile source clean
+.PHONY: source clean
 
 source:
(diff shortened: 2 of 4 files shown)


st3/sublime_lib/settings_dict.py
```diff
@@ -181,10 +182,7 @@ def onchange():
 
         key = str(uuid4())
         self.settings.add_on_change(key, onchange)
-        return key
-
-    def unsubscribe(self, key):
-        self.settings.clear_on_change(key)
+        return lambda: self.settings.clear_on_change(key)
```
FichteFoll: I usually prefer `functools.partial` over lambdas because they are more robust, though in this particular case there shouldn't be any problems.

FichteFoll:
I agree that returning the generated key is pretty much useless, but I also don't think returning a function to clear the callbacks is really an improvment.

Instead, I would rather have the `unsubscribe` method use the same key that was generated in `subscribe` earlier and stored in an attribute, because you're much more likely to keep the DictSettings instance around rather than having to take care of yet another potentially global variable in the unsubscribe header. It also reduces the number of variables.

FichteFoll:
>I'm not sure what you mean by 'stored in an attribute'. Do you mean an attribute of SettingsDict?

Yes.

>Client code could subscribe several times to one SettingsDict object

Ah well, I forgot about that. That does indeed make things quite different as the only way my suggestion would work is to store all generated keys in a set and then unsubscribe from all at once. Not ideal if you don't want that.

>two SettingsDict objects wrapping the same Settings are interchangeable for all purposes.

Would you say this is a required feature? It does exist currently, but I never thought of using `SettingsDict` that way, especially because when you provide some defaults in the constructor, for example, and then wish to keep that instance around, where it actually *does* have state, albeit one that doesn't change after initializing.
It just seems to me that the most prominent use case is to share a `SettingsDict` instance across your code that you can rely on, that you will create in `plugin_loaded`, specify defaults, add subscriptions, and then unregister any subscriptions in `plugin_unloaded`. If you need to keep the keys or unsubscribe handles around (which you do in case the plugin gets unloaded), you might as well just keep the settings instance around.

The usage you describe doesn't make sense to me. I do understand the use case of only wanting to unsubscribe for a single callback (for properties), however, and for that you'd need a way to differentiate between them...

Thom1729:
> Would you say this is a required feature?

Not required, just a nice bonus — and possibly avoiding a subtle “gotcha”. My main concern is about the `SettingsDict` having two kinds of state, one kind shared between instances and one kind unique per instance.

> especially because when you provide some defaults in the constructor, for example

We removed the defaults in #38 in favor of `collections.ChainMap`.

My thought process is basically that unless we forbid multiple independent subscriptions (which the underlying API is designed to allow), then we basically have to return some unique value from `subscribe` for the client to keep track of. The most opaque, self-contained value we could return is a callback that unsubscribes. Any other value, such as a key, would require that the client keep track of two things instead of one.

I concur that it'll be pretty common for packages to keep a `SettingsDict` object around. That's probably what I'll be doing in JS Custom. In that case, it doesn't make a difference whether you need the original `SettingsDict` to unsubscribe. That isn't really an upside, though; it's a wash unless you run into the “gotcha”.

FichteFoll:
>We removed the defaults in #38 in favor of collections.ChainMap.

Aw, I feel like I've been found out. I mentioned this before, but I haven't been in touch with developing new plugins in quite a while and thus have never used sublime_lib myself. Kind of a bad premise for reviewing, but I suppose it kind of worked out so far.

Anyway, I guess we can merge this as-is now, since the previous behavior wasn't documented, but remember that changing it *after* a release is a major change.
Currently, I believe a good way forward is probably to support both usage patterns, i.e. return a way to unsubscribe from invidivual callbacks but also track all subscriptions in the instance and then provide a convenience method to unsubscribe from all of them at once. I will see when I run into the situation of needing it and probably suggest a change myself then.

PR #48, opened 2018-11-05
Fixed indentation for print

Thom1729:
Apparently the `ViewStream.print` function wasn't handling indentation correctly. I feel like it should have worked, but it didn't. Now it does.

This PR just implements `print` directly rather than using the builtin. The docstring is probably more explicit like this anyway. As a side effect, it will raise `ArgumentError` if the user passes `file` or `flush` arguments.

### DIFF
--- a/tests/test_view_stream.py
+++ b/tests/test_view_stream.py
@@ -127,11 +127,16 @@ class TestViewStream(TestCase):
         self.assertContents('')
 
     def _compare_print(self, *args, **kwargs):
+        s = StringIO()
+        print(*args, file=s, **kwargs)
+
         self.stream.clear()
         self.stream.print(*args, **kwargs)
 
-        s = StringIO()
-        print(*args, file=s, **kwargs)
+        self.assertContents(s.getvalue())
+
+        self.stream.clear()
+        print(*args, file=self.stream, **kwargs)
 
         self.assertContents(s.getvalue())
 
--- a/tests/test_view_stream.py
+++ b/tests/test_view_stream.py
@@ -110,17 +110,21 @@ class TestViewStream(TestCase):
         self.assertEqual(text, "World")
         self.assertEqual(self.stream.tell(), 12)
 
-    def test_write_read_only(self):
+    def test_write_read_only_failure(self):
         self.stream.view.set_read_only(True)
 
         self.assertRaises(ValueError, self.stream.write, 'foo')
         self.assertRaises(ValueError, self.stream.clear)
 
+    def test_write_read_only_success(self):
+        self.stream.view.set_read_only(True)
         self.stream.force_writes = True
-        self.assertEqual(self.stream.write('foo'), 3)
+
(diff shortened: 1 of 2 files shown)


FichteFoll:
That's weird if `print(file=ViewStream)` doesn't work as intended. It's also significant becase we'd want to support the file protocol (it's used by various utilities in the stdlib and otherwise). Can you share some details?

The particular change seems okay, but I don't get the why.

Thom1729:
After some investigation, the following test fails:

```python
    def test_no_indent(self):
        text = "    "

        self.stream.write(text)
        self.stream.write('\n')
        self.assertContents(text+'\n')
```

It looks like the `insert_snippet` trick isn't working. After `text` is inserted, the line has a four-space indent. When '\n' is inserted, another indent is inserted after the newline. This happens regardless of the `auto_indent` setting. However, the following test succeeds:

```python
    def test_no_indent(self):
        text = "    "

        self.stream.write(text+'\n')
        self.assertContents(text+'\n')
```

This has to be a quirk of `insert_snippet`. It looks like to handle indentation correctly, we'll have to use `insert` and manage the `auto_indent` setting (much like we manage the read-only property). Once that's done, the change to `print` will be unnecessary.

Thom1729:
I've rewritten `write` to solve the indentation issue (hopefully) once and for all. I'm not a big fan of patching/unpatching the `auto_indent` setting, but it's necessary in order to preserve the originally intended behavior.

Tests have been added/updated. They will now fail if you remove the `@guard_auto_indent` decorator.

I think that explicit args for `print` still make sense, and that the docs are better with them. The implementation has been reverted to use the builtin again.

tests/test_view_stream.py
```diff
@@ -121,6 +122,43 @@ def test_write_only(self):
...
+        self.stream.print(text)
+        self.assertContents(text + "\n")
+
+    def test_print_read_only(self):
+        self.stream.view.set_read_only(True)
+
+        self.assertRaises(ValueError, self.stream.print, 'foo')
+        self.assertRaises(ValueError, self.stream.clear)
+
+        self.stream.force_writes = True
```
FichteFoll: I believe this belongs to a separate test case.

FichteFoll:
All right, this looks good. The fact that there is no way to prevent `auto_indent` from interfering when inserting something in the view is quite annoying, but at least we found a functioning workaround.

Left a comment regarding a test, otherwise this is fine to merge (i.e. squash-merge).

tests/test_view_stream.py
Thom1729: If so, it should be changed in `test_write_read_only` as well.

FichteFoll: Yes, probably. That test case wasn't included in the diff.

Thom1729:
In addition to splitting the read-only tests, I slightly rewrote them to use `assertContents`. I also improved the print tests to compare against `print(file=self.stream)`.

Issue #31, opened 2018-05-05
Wrapper for `Window.show_quick_panel` to choose an item from a list.


Thom1729:
The most common use for `Window.show_quick_panel` is to choose one item from a list of items, then do something with the result. The API for the method is a bit inconvenient. You pass in a list of labels, and the callbacks get a numerical index. Most of the time, that index is not particularly meaningful; you have to index your original list to find out what it's for (after checking that it isn't `-1`). This makes for a fair bit of boilerplate in this common case:

```python
syntaxes = [
    s for s in sublime_lib.syntaxes.list_syntaxes()
    if not s.hidden
]

labels = [ s.name for s in syntaxes ]

def on_selected(index):
    if index == -1:
        return

    syntax = syntaxes[index]
    do_something_with_syntax(syntax)

window.show_quick_panel(
    items=labels,
    on_done=on_selected
)
```

This example is (reconstructed from memory) from a syntax selection package I'm working on, but I've seen this pattern many places. Instead, I suggest a utility function providing a simpler interface inspired by typical HTML `<select>` widgets:

```python
show_selection_panel(
    window,
    items=[
        s for s in sublime_lib.syntaxes.list_syntaxes()
        if not s.hidden
    ],
    labels=lambda s: s.name,
    on_selected=do_something_with_syntax,
)
```

The API might look something like:

```python
def show_selection_panel(
    window,
    items, # `list` of items of whatever type
    *,
    flags=None,
    labels=str, # Function producing a label from an item
    selected=NOT_GIVEN, # An item, not an index
    on_selected=None, # When an item is chosen.
    on_cancel=None, # When the panel is cancelled.
    on_highlighted=None, # When an item is highlighted. Receives the item.
):
```

Does this seem like a good, generally useful function to provide?

keith-hall:
Sounds useful to me - in some cases it's annoying to have to "store" the list somewhere manually just to know which item was picked, so having `sublime_lib` take care of it would be awesome.

FichteFoll:
(I don't think I've ever seen a single package using `View.show_popup_menu`. :thinking: )

PR #50, opened 2018-11-09
Added show_selection_panel.

Thom1729:
For #31.

I think the function is fine, or close to it. The tests, on the other hand, are a bit gnarly.

### DIFF
--- a/st3/sublime_lib/show_selection_panel.py
+++ b/st3/sublime_lib/show_selection_panel.py
@@ -20,7 +20,8 @@ def show_selection_panel(
 
     Optional keyword arguments:
 
-    :argument flags: TODO
+    :argument flags: A bitwise OR of :const:`sublime.MONOSPACE_FONT` and
+    :const:`sublime.KEEP_OPEN_ON_FOCUS_LOST`.
 
     :argument labels: A function taking elements of `items` to string labels.
 
--- a/st3/sublime_lib/show_selection_panel.py
+++ b/st3/sublime_lib/show_selection_panel.py
@@ -1,4 +1,14 @@
+from collections.abc import Sequence
+
+
+__all__ = ['show_selection_panel']
+
+
 NOT_GIVEN = object()
+
+
+def is_sequence_not_str(obj):
+    return isinstance(obj, Sequence) and not isinstance(obj, str)
 
 
 def show_selection_panel(
@@ -16,8 +26,7 @@ def show_selection_panel(
 
     :argument window: The :class:`sublime.Window` in which to show the panel.
 
-    :argument items: A :class:`list` of items to choose from. :exc:`ValueError`
-    will be raised if `items` is empty.
+    :argument items: A nonempty :class:`list` of items to choose from.
 
     Optional keyword arguments:
 
@@ -25,13 +34,12 @@ def show_selection_panel(
     :const:`sublime.KEEP_OPEN_ON_FOCUS_LOST`.
 
     :argument labels: Either a list of labels or a function taking elements of
-    `items` to string labels. If `labels` is not given, the `items` themselves
-    will be taken as the `labels`.
+    `items` to labels. If `labels` is not given, it will default to `items`.
 
-    Every label must be either a single value or a list of values.
-    :exc:`ValueError` will be raised if some labels are lists and others are not
-    or if the labels are lists with inconsistent lengths. Each value will be
-    converted to a string via :func:`str`.
+    Every label must be a single item (a string or convertible with :func:`str`)
+    or a :class:`collections.abc.Sequence` of items (such as a :class:`list` or
+    :class:`tuple`). In the latter case, each entry in the quick panel will show
+    multiple rows.
 
     :argument selected: The value in `items` that will be initially selected.
     If `selected` is not given, no value will be initially selected.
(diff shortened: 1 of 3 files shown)


st3/sublime_lib/show_selection_panel.py
```diff
@@ -0,0 +1,68 @@
+NOT_GIVEN = {}
```
FichteFoll:
```suggestion
NOT_GIVEN = object()
```

st3/sublime_lib/show_selection_panel.py
```diff
@@ -0,0 +1,68 @@
...
+    :argument on_highlight: A callback accepting a value from `items` to be
+    invoked every time the user changes the highlighted item in the panel.
+    """
+    labels_list = list(map(labels, items))
+
+    def on_done(index):
+        if index == -1:
+            if on_cancel:
+                on_cancel()
+        else:
```
FichteFoll:
```suggestion
        elif on_select:
```

And remove the next line and unindent the following. No idea how to submit multi-line suggestions.

st3/sublime_lib/show_selection_panel.py
```diff
@@ -0,0 +1,68 @@
...
+            if on_select:
+                on_select(items[index])
+
+    if selected == NOT_GIVEN:
```
FichteFoll:
```suggestion
    if selected is NOT_GIVEN:
```

st3/sublime_lib/show_selection_panel.py
```diff
@@ -0,0 +1,68 @@
...
+    :argument window: The :class:`sublime.Window` in which to show the panel.
+
+    :argument items: A :class:`list` of items to choose from.
+
+    Optional keyword arguments:
+
+    :argument flags: A bitwise OR of :const:`sublime.MONOSPACE_FONT` and
+    :const:`sublime.KEEP_OPEN_ON_FOCUS_LOST`.
+
+    :argument labels: A function taking elements of `items` to string labels.
```
FichteFoll: `items` (the parameter to `show_quick_panel`) may be a list of strings for multi-line display. Although we don't *have* to support it, it'd be a waste imo. So, a "label" can also be a list of strings, but their length should be consistent due to ST bugs.

Furthermore, I suggest an alternative use case where you already have the labels to show in the panel and the values to resolve to in a collection (or two). A list of two-tuples (or just two-item collections) sounds like a good compromise, where you'd have a list like follows:
```py
items = [
    ("label1", "value1"),
    ("label2", "value2"),
]
```
Allowing labels to be a list would be an easy way for `labels, values = zip(*items)` and then using these two lists in the `show_selection_panel` call.
Alternatively, passing a value->label map here might also seem useful, but that'd need some exploration. All values must be hashable in that case.

Also not so fond of the parameter name, if only functions are accepted. `labels` seems to imply it would expect a collection of labels, not a function that maps items to labels.

TL;DR
- Allow a list to be passed as the labels.
- Still allow a function to be passed. Useful for the default value of `str`, for example.
- Explore allowing a dict to be passed that has both values and labels; still need a separate param for order? value->label or label->value? different param?

st3/sublime_lib/show_selection_panel.py
```diff
@@ -0,0 +1,94 @@
...
+        raise ValueError("The items parameter must contain at least one item.")
+
+    if labels is NOT_GIVEN:
+        labels = items
+    elif callable(labels):
+        labels = list(map(labels, items))
+    elif len(items) != len(labels):
+        raise ValueError("The lengths of `items` and `labels` must match.")
+
+    if any(isinstance(label, list) for label in labels):
```
FichteFoll: What about allowing any sized collection that's not a string instead? (These are the situations where I wish strings weren't iterable.) Tuples satisfy that criteria, sets don't.

When tuples are allowed we can also used those in the various `tuple(map(...))` calls within this function, as tuples have less overhead due to being immutable.

st3/sublime_lib/show_selection_panel.py
```diff
@@ -0,0 +1,94 @@
...
+
+    :argument on_select: A callback accepting a value from `items` to be
+    invoked when the user chooses an item.
+
+    :argument on_cancel: A callback that will be invoked with no arguments if
+    the user closes the panel without choosing an item.
+
+    :argument on_highlight: A callback accepting a value from `items` to be
+    invoked every time the user changes the highlighted item in the panel.
+    """
```
FichteFoll: Add `:raises:` to docstrings? Doesn't need to be as detailed as to include every individual exception, but should at least mention all exception types. Also consider the exception that `list.index` throws.

st3/sublime_lib/show_selection_panel.py
```diff
@@ -0,0 +1,94 @@
...
+    elif callable(labels):
+        labels = list(map(labels, items))
+    elif len(items) != len(labels):
+        raise ValueError("The lengths of `items` and `labels` must match.")
+
+    if any(isinstance(label, list) for label in labels):
+        if not all(isinstance(label, list) for label in labels):
+            raise ValueError("Labels must be all strings or all lists.")
+
+        if len(set(map(len, labels))) != 1:
```
FichteFoll: I like this :+1:

FichteFoll:
So yeah, no further changes requested but just a few things to think about.

Not sure when I'll manage to review your other PR. Maybe later today, maybe tomorrow.

st3/sublime_lib/show_selection_panel.py
Thom1729: The underlying API takes lists, so we should pass those, but handling tuples ourselves is definitely a good idea.

PR #52, opened 2018-11-14
Added ResourcePath.

### LINKED ISSUE #25: Utilities for working with Sublime resource paths
Many Sublime features and API functions deal with resource paths, such as `Packages/JavaScript.JavaScript.sublime-syntax`. On the other hand, OS features require "real" system paths. When working with Sublime resources, one often must keep track of both the Sublime resource path and the system path.

For example, when running JS Custom's syntax tests, I need to:

- Locate the test directory.
- Clear/create it (by system path).
- Compile a test syntax into the test directory (by system path).
- Copy a syntax test file into the test directory (by system path).
- Modify the syntax test (by system path) to refer to the compiled test syntax file (by resource path).
- Tell the Sublime API to run the syntax test (by resource path).

Keeping separate track of the resource and system paths to the several files involved is annoying, verbose, and error-prone.

Another example is the suggested resource caching system in #24. The cache takes requests for sublime resources, but to implement the caching it also needs the system file path and last modification time. Other packages that want to work with resources in more sophisticated ways, such as [OverrideAudit](https://github.com/OdatNurd/OverrideAudit/) and [PackageResourceViewer](https://github.com/skuroda/PackageResourceViewer/), have to solve these problems manually.

A convenient general-purpose system for dealing with resources could be very useful to package developers, helping them to solve some otherwise tricky problems while getting the corner cases right. As you may have guessed, I'm writing this here because I think that we can provide such a system.

The lightweight solution might be a handful of utility functions to convert between resource and system paths. We might also provide functions to (for example) get the last modification time of a resource.

A more comprehensive solution would be to define a SublimeResource type inspired by [pathlib](https://docs.python.org/3/library/pathlib.html). As with pathlib, users could perform all of the common file operations on a SublimeResource -- reading text or binary data, listing directory contents, and so on. These operations would abstract away the underlying complexity of the resource system. For instance, `read_bytes` and `read_text` would be implemented using `sublime.load_binary_resource` and `sublime.load_resource`, whereas `iterdir` would be implemented using both `os.listdir` and `ZipFile.namelist`. In addition, SublimeResource would provide Sublime-specific functionality, like extracting the package name. Ideally, we could provide a `Path`-compatible interface, though a few methods might not be supportable.

Thoughts?

Thom1729:
Closes #25.

Feedback requested.

### Background

Sublime's resource system resembles a filesystem in some ways but not in others. The base functionality consists of `sublime.find_resources`, which provides a list of paths, and `sublime.load_(binary)_resource`, which given a path returns the value of the resource. These paths resemble filesystem paths in many ways, and it makes sense to represent them an abstraction similar to `pathlib`.

The contents of the resource system -- resource paths and values -- are determined by the filesystem. However, there are several important caveats:

- A resource may represent a file in the filesystem, or it may represent the partial contents of a zip file rather than an entire individual file.
- The resource system has no concept of directories. In the filesystem, the parent of a path to a file or directory must point to a directory. In the resource system, the parent of a path to a resource may refer to a resource or to nothing at all.
- The resource system is immutable (through Sublime's interface). You can find and load resources, but not save them or create new ones. Instead, you would directly modify the filesystem under `sublime.packages_path()`. (`sublime.save_settings` does affect the resource system, but I see this as an indirect side effect.)
- The resource system has no concept of a working directory or of path traversal. This makes the concept of relative paths dubious.
- The resource system doesn't have symlinks, drives, permissions, stats, devices, mount points, or other such complications. Everything that objectively exists is a resource.

Because of all these differences, I don't think it makes sense to pretend that a resource path is a filesystem path (e.g. extending `pathlib.Path`). The two are similar and closely related, but they refer to different kinds of things, and I think that conflating them would lead to awkward compromises.

### Functionality

A resource path abstraction could help with tasks like the following:

- Building and manipulating resource paths.
- Finding all resources of a certain description in a certain package.
- Creating or modifying files in the Packages directory corresponding to a certain resource path.

### Proposal

This PR adds two new classes, `PureResourcePath` and `ResourcePath`. `PureResourcePath` provides basic path manipulation features very similar to `PurePath`. `ResourcePath` (a subclass of `PureResourcePath`) also interfaces with the Sublime API to find and load resources and with `pathlib` to provide usable `Path` objects.

#### PureResourcePath

An instance of `PureResourcePath` wraps a tuple of path components, implementing most of the functionality of `PurePath`.

- There are no relative paths. `is_absolute()` and `relative_to()` are omitted.
- Irrelevant compatibility features are excluded: `drive`, `anchor`, `as_posix()`, `as_uri()`, `is_reserved()`.
- `PureResourcePath` does not depend on `sublime` and could in principle be used in any environment.
- `PureResourcePath.root` returns the first path component, which should generally be `Packages` or `Cache`.
- `PureResourcePath.package` returns the second path component.

#### ResourcePath

`ResourcePath`, by contrast, does not implement most of the functionality of `Path`.

- `exists` returns True if the `ResourcePath` corresponds to a resource. (Directories are not resources.)
- `read_text` and `read_bytes` wrap `sublime.load_resource` and `sublime.load_binary_resource`.
- `glob` works much like `Path.glob`, but it never returns directories (because directories are not resources). `rglob`, likewise.
- `children` "morally" lists directory contents. It returns all direct child paths that contain resources at or below them. `exists` may return False for some children!
- `file_path` returns the `Path` of the matching filesystem location inside `sublime.packages_path()` or `sublime.cache_path()`. It does so regardless of whether that `Path` exists in the filesystem. If `pathlib` is not available, it returns a stub object that should play well with `str` and `os.fspath`.

### Usage

The JS Custom example in #25 could be implemented as follows:

```py
test_dir = ResourcePath('Packages/User/JS Custom/tests')
test_dir.file_path().mkdir(parents=True, exist_ok=True)

syntax_path = test_dir / 'example.sublime-syntax'
yamlmacros.compile(some_stuff, to=syntax_path.file_path()) # Actual call elided.

syntax_test_sources = ResourcePath('Packages/JSCustom/tests/syntax_test_example.js').glob('syntax_test_*.js')
for source in syntax_test_sources:
    dest_path = test_dir / source.name
    with dest_path.file_path().open('w') as dest:
        dest.write('// SYNTAX TEST "%s"\n' % syntax_path)
        dest.write(source.read_text())

    sublime_api.run_syntax_test(dest_path)
```

#### Why not implement `ResourcePath.write`, etc?

Sublime's API doesn't provide mutation. When reading a resource, we don't care about what's backing it in the filesystem (a real file/a zipped file; where that file is located), but when mutating a resource it makes a big difference. In most cases, the sensible behavior would be to create a new file/modify an existing file in Packages (or Caches). In these cases, we can do the mutation through `file_path()` and the meaning is clear and explicit. Alternatively, we could mutate an existing "file" instead of creating a new one, even if the existing file is inside a zip archive, but that's a can of worms I don't think we need.

This is a point that tripped me up when I first considered #25. I had an idea to make `ResourcePath` work like a `ChainMap`. But in this case, I think the abstraction would be leaky and overcomplicated. The cost to omitting mutation is an extra `.file_path()` every now and then, and I think that's acceptable. We can always add more features if they make sense later.

#### Should we bundle `pathlib`?

Right now, `ResourcePath.file_path()` returns a `pathlib.Path` if that dependency is available or a stub `Path` if it isn't. The idea is that a user who wants the features of `pathlib` can declare the dependency themselves, whereas a user who doesn't care can just call `str` on the stub (or pass it to a sane implementation of `os.fspath`). This might be a bit hacky.

If we could depend on another dependency, I'd say that we should depend on `pathlib`. As-is, we could vendor it. Like `enum`, we presumably wouldn't expose it. (This solution mildly annoys me, but there's nothing really wrong with it.)

We could also simply return a string from `file_path`, but this would make adding `pathlib` later a breaking change. We're all anxiously hoping for a newer Python runtime real-soon-now, so `pathlib` seems like the way to go. I also think that omitting `file_path()` is not an option; it's arguably the keystone of the feature.

#### Other

- A `ResourcePath.copy_to(actual_filesystem_path)` method could be useful, especially to copy a resource "tree".
- The `glob` implementation is not full-featured. We probably want to match `pathlib`'s.
- `glob` could probably be optimized to pass the last part of the pattern to `sublime.find_resources`. This might not be compatible with added `glob` features.
- This PR does not refer to `sublime.installed_packages_path()` at all. I'm not sure what features, if any, we might add for working with installed packages. It may be that a user working with installed packages would want a different abstraction anyway.
- This PR still needs docs and probably more tests.
- The idea of `ResourcePath.root` being `Packages` or `Cache` makes sense for Sublime, but this implementation isn't perfectly consistent about it. If those directories represent roots, then we shouldn't allow an empty `ResourcePath`. If the "real" root is the empty path, though, then we should rename the property that returns `self.parts[0]`.


st3/sublime_lib/glob_util.py
```diff
@@ -0,0 +1,21 @@
...
+def get_glob_expr(pattern):
+    s = ''
+    for part in GLOB_RE.split(pattern):
+        if part == '**/':
+            s += r'(?:.*/)?'
+        elif part == '**':
+            s += r'(?:.*)'
+        elif part == '*':
+            s += r'(?:[^/]*)'
+        else:
```
FichteFoll: What about `?` characters? They should probably be converted to `.`.

I suppose we can ignore `[set]` matching.

st3/sublime_lib/path_compat.py
```diff
@@ -0,0 +1,22 @@
+__all__ = ['Path']
```
FichteFoll: A docstring for the motivation of this module would be appreciated.

Also maybe rename it to `_path_compat` to signal that it's supposed to private? Although users *might* potentially want to access the class type :thinking:

st3/sublime_lib/pure_resource_path.py
```diff
@@ -0,0 +1,103 @@
+import posixpath
+from functools import total_ordering
+
+
+__all__ = ['PureResourcePath']
+
+
+@total_ordering
+class PureResourcePath():
+    """A representation of a Sublime Text resource path. Inspired by pathlib."""
```
FichteFoll: Should mention that only `/` are supported, not `\`.

And that not all methods of pathlib.PurePath are implemented, but those that make sense.

st3/sublime_lib/pure_resource_path.py
```diff
@@ -0,0 +1,103 @@
...
+            part
+            for segment in pathsegments if segment
+            for part in posixpath.normpath(str(segment)).split('/')
+        )
+
+    def __hash__(self):
+        return hash(self.parts)
+
+    def __repr__(self):
+        return "%s(%s)" % (type(self).__name__, repr(str(self)))
```
FichteFoll: Can also use `self.__class__` instead of `type(self)`. No idea about the benefits but I always use the former.

```suggestion
        return "{}({!r})".format(self.__class__.__name__, str(self))
```

st3/sublime_lib/pure_resource_path.py
```diff
@@ -0,0 +1,103 @@
...
+
+    def __str__(self):
+        return '/'.join(self.parts)
+
+    def __eq__(self, other):
+        return self._parts == other.parts
```
FichteFoll: Should be checking other's type and return `NotImplemented` if it's not an instance of `PureResourcePath`.

st3/sublime_lib/pure_resource_path.py
```diff
@@ -0,0 +1,103 @@
...
+
+    def __lt__(self, other):
```
FichteFoll: Same as for `__eq__`, though I wonder whether we should implement this at all. I mean, how do you compare paths? Though, pathlib provides them, so we should follow that.

st3/sublime_lib/pure_resource_path.py
```diff
@@ -0,0 +1,103 @@
...
+        return self._parts < other.parts
+
+    def __truediv__(self, other):
+        return self.joinpath(other)
+
+    def _new(self, *args):
```
FichteFoll: Remove this and replace usages with `self.__class__(*args)` imo.

st3/sublime_lib/pure_resource_path.py
```diff
@@ -0,0 +1,103 @@
...
+
+    @property
+    def parent(self):
+        if len(self._parts) > 1:
+            return self._new(*self._parts[:-1])
+        else:
+            return self
+
+    @property
+    def parents(self):
```
FichteFoll: `pathlib.parents` returns an *immutable* sequence, so this should be a tuple.

st3/sublime_lib/pure_resource_path.py
```diff
@@ -0,0 +1,103 @@
...
+            return []
+        else:
+            return [parent] + parent.parents
+
+    @property
+    def name(self):
+        try:
+            return self._parts[-1]
+        except IndexError:
+            return None
```
FichteFoll: Should return an empty string.
```suggestion
            return ''
```
```
In [20]: PurePath("").name
Out[20]: ''
```

This property is also used in various places and expected to be a string.

st3/sublime_lib/pure_resource_path.py
```diff
@@ -0,0 +1,103 @@
...
+    def suffix(self):
+        return posixpath.splitext(self.name)[1]
+
+    @property
+    def suffixes(self):
+        return ['.' + suffix for suffix in self.name.split('.')[1:]]
+
+    @property
+    def stem(self):
+        return self.name.split('.')[0]
```
FichteFoll:
```suggestion
        return posixpath.splitext(self.name)[0]
```

Could also use rsplit, but if we import `posixpath` we might as well use it for consistency.

st3/sublime_lib/pure_resource_path.py
```diff
@@ -0,0 +1,103 @@
...
+
+    @property
+    def root(self):
+        try:
+            return self._parts[0]
+        except IndexError:
+            return None
```
FichteFoll:
```suggestion
            return ''
```
```
In [20]: PurePath("").name
Out[20]: ''
```

st3/sublime_lib/pure_resource_path.py
```diff
@@ -0,0 +1,103 @@
...
+
+    @property
+    def package(self):
+        try:
+            return self._parts[1]
```
FichteFoll: We never verify that the resource path points to `Packages/*`.

st3/sublime_lib/pure_resource_path.py
```diff
@@ -0,0 +1,103 @@
...
+        except IndexError:
+            return None
+
+    def joinpath(self, *other):
+        return self._new(self, *other)
+
+    def with_name(self, name):
+        if not self.name:
+            raise ValueError("%r has an empty name" % (self,))
```
FichteFoll:
```suggestion
            raise ValueError("{!r} has an empty name".format(self))
```
Please alwas use `str.format` over `%` formatting. I mean, it's your code base, but I vastly prefer that style, especially because of single-element tuples requiring that trailing comma.

st3/sublime_lib/pure_resource_path.py
```diff
@@ -0,0 +1,103 @@
...
+
+        return self.parent / name
+
+    def with_suffix(self, suffix):
+        return self.parent / (posixpath.splitext(self.name)[0] + suffix)
```
FichteFoll: `relative_to` and `match` are not implemented.

A couple other methods and properties of PurePath aren't as well, but they don't make sense to have on a resource path. Should we raise `NotImplementedError` for those?

st3/sublime_lib/resource_path.py
```diff
@@ -0,0 +1,83 @@
+import sublime
+import os.path
+from collections import OrderedDict
+
+from .pure_resource_path import PureResourcePath
```
FichteFoll: I would actually define this class in the same file and remove `pure_resource_path`. I don't believe separating the classes makes sense when they are basically the same thing.

st3/sublime_lib/resource_path.py
```diff
@@ -0,0 +1,83 @@
...
+
+    def children(self):
+        depth = len(self._parts)
+        return [
+            self / next_part
+            for next_part in OrderedDict.fromkeys(
+                resource.parts[depth]
+                for resource in self.glob('**')
+            )
+        ]
```
FichteFoll: See `NotImplementedError` remark on `PureResourcePath`. Though with how many methods there are, this would get messy rather quickly. I'm in favor of just not defining the methods.

tests/test_pure_resource_path.py
```diff
@@ -0,0 +1,186 @@
...
+        self.assertEqual(
+            str(ResourcePath("Packages/Foo/bar.py")),
+            "Packages/Foo/bar.py"
+        )
+
+    def test_repr(self):
+        self.assertEqual(
+            repr(ResourcePath("Packages/Foo/bar.py")),
+            "ResourcePath('Packages/Foo/bar.py')"
+        )
```
FichteFoll: Testing the edge case `ResourcePath("")` would probably be useful.

tests/test_pure_resource_path.py
```diff
@@ -0,0 +1,186 @@
...
+    def test_parent(self):
+        self.assertEqual(
+            ResourcePath("Packages/Foo/bar.py").parent,
+            ResourcePath("Packages/Foo")
+        )
+
+    def test_top_parent(self):
+        self.assertEqual(
+            ResourcePath("Packages").parent,
+            ResourcePath("Packages")
```
FichteFoll: Why? Shouldn't this give `ResourcePath("")`?
The root is the data dir after all.

```
In [28]: PurePath("abc").parent
Out[28]: PurePosixPath('.')
```

tests/test_pure_resource_path.py
```diff
@@ -0,0 +1,186 @@
...
+    def test_stem(self):
+        self.assertEqual(
+            ResourcePath("Packages/Foo/bar.py").stem,
+            'bar'
+        )
+
+    def test_stem_multiple(self):
+        self.assertEqual(
+            ResourcePath("Packages/Foo/bar.tar.gz").stem,
+            'bar'
```
FichteFoll: As I mention in the implementation of `stem`, this is wrong and should return `bar.tar`.
```suggestion
            'bar.tar'
```

```
In [29]: PurePath("a.b.c").stem
Out[29]: 'a.b'
```

tests/test_pure_resource_path.py
```diff
@@ -0,0 +1,186 @@
...
+    def test_root(self):
+        self.assertEqual(
+            ResourcePath("Packages/Foo/bar").root,
+            'Packages'
+        )
+
+    def test_root_none(self):
+        self.assertEqual(
+            ResourcePath("").root,
+            None
```
FichteFoll: Hm, I guess we should be returning `""` here rather.

```
In [30]: PurePath("").root
Out[30]: ''
```

tests/test_resource_path.py
```diff
@@ -0,0 +1,90 @@
...
+        with self.assertRaises(FileNotFoundError):
+            ResourcePath("Packages/test_package/nonexistentfile.txt").read_bytes()
+
+    def test_read_bytes_invalid_unicode(self):
+        # Should not raise UnicodeDecodeError
+        ResourcePath("Packages/test_package/UTF-8-test.txt").read_bytes()
+
+    def test_glob(self):
+        self.assertEqual(
+            ResourcePath("Packages/test_package").glob('*.txt'),
```
FichteFoll: Should test `**` (and `?`) as well.

st3/sublime_lib/resource_path.py
```diff
@@ -0,0 +1,83 @@
...
+        return glob_resources(base + pattern)
+
+    def rglob(self, pattern):
+        return self.glob('**/' + pattern)
+
+    def children(self):
+        depth = len(self._parts)
+        return [
+            self / next_part
+            for next_part in OrderedDict.fromkeys(
```
FichteFoll: Why iterate over an ordered dict created with fromkeys? Might as well just use a normal generator, which is also ordered.

```py
return [self / resource.parts[depth] for resource in self.glob('**')
```

Actually, why not just use `self.glob('*')`?

Actually-actually, why have this in the first place?

st3/sublime_lib/pure_resource_path.py
```diff
@@ -0,0 +1,103 @@
...
+    def _new(self, *args):
+        return type(self)(*args)
+
+    @property
+    def parts(self):
+        """A tuple giving access to the path’s various components."""
+        return self._parts
+
+    @property
+    def parent(self):
```
FichteFoll: No docstrings beyond this point.

st3/sublime_lib/path_compat.py
```diff
@@ -0,0 +1,22 @@
...
+    from pathlib import Path
+except ImportError:
+    import os.path
+
+    class Path():
+        def __init__(self, path):
+            self.text = os.path.normpath(path)
+
+        def __repr__(self):
+            return "%s(%s)" % (type(self).__name__, repr(self.text))
```
FichteFoll:
```suggestion
            return "{}({!r})".format(self.__class__.__name__, self.text)
```

(See below.)

st3/sublime_lib/pure_resource_path.py
Thom1729: Done.

st3/sublime_lib/pure_resource_path.py
Thom1729: Following the example of `pathlib`, it now returns `False` for mismatched types.

st3/sublime_lib/pure_resource_path.py
Thom1729: Now raises `TypeError`.

st3/sublime_lib/pure_resource_path.py
Thom1729: Done.

st3/sublime_lib/pure_resource_path.py
Thom1729: Done.

st3/sublime_lib/pure_resource_path.py
Thom1729: Done.

st3/sublime_lib/pure_resource_path.py
Thom1729: Done.

st3/sublime_lib/pure_resource_path.py
Thom1729: It could also point to `Cache/*`. I'm not sure what else we can say about resource paths that don't begin with one `Packages` or `Caches`.

st3/sublime_lib/resource_path.py
Thom1729: This method gets all immediate child paths that contain a resource or are the prefix of a path that contains a resource. If this were a filesystem, it would represent files and directories, but because resources have no concept of directories, the interpretation is slightly different and the implementation is substantially different.

`OrderedDict.fromkeys` gets unique values in order.

st3/sublime_lib/path_compat.py
FichteFoll: I agree that the stub is suboptimal. It tries to be a pathlib.(Pure)Path but is not. We might as well just return a string if pathlib isn't available, so requiring/vendoring pathlib seems like the better approach.

I suppose until dep-deps are implemented, we won't get around vendoring to ensure proper and predictable functionality.

st3/sublime_lib/pure_resource_path.py
```diff
@@ -48,13 +48,16 @@ def parts(self):
 
     @property
     def parent(self):
+        """The logical parent of the path:"""
```
FichteFoll: This docstring seems unfinished.

st3/sublime_lib/pure_resource_path.py
FichteFoll: Or not?

st3/sublime_lib/pure_resource_path.py
FichteFoll: We could also check if the first part is `Packages` first and only then return the second part. That way it would always return the package's name or an empty string.

For all other purposes there is the `parts` property.

st3/sublime_lib/pure_resource_path.py
FichteFoll: Agreed. `relative_to` also doesn't make sense because resource paths are always absolute. Please resolve when read (not sure if you'd catch it otherwise).

st3/sublime_lib/pure_resource_path.py
```diff
@@ -103,13 +114,20 @@ def match(self, pattern):
...
+        """Return a new path with the name changed.
+
+        :raise ValueError: if the original path doesn’t have a name."""
         if not self.name:
             raise ValueError("{!r} has an empty name".format(self))
 
         return self.parent / name
 
     def with_suffix(self, suffix):
+        """Return a new path with the suffix changed. If the original path
```
FichteFoll: Should begin a new paragraph after the first sentence (for the short description).

tests/test_pure_resource_path.py
FichteFoll: I guess this depends on our definition. If we think of resource paths more like Windows paths with the drive names as roots, it would make sense to return the first part on `.root` and to not go beyond single-part paths.

I am fine with either, both have pros and cons. However, we should document what we think of when talking about the "root" and .parent and .root should have aligning behavior.

st3/sublime_lib/resource_path.py
FichteFoll: Ah, I get it now. `chilren` includes paths that would be directories but don't (really) exist in resource paths. I suggest we add a method that tests whether a resource path would point to a directory (by checking whether there are paths that match `str(dir_path) + '/**'`).

FichteFoll:
What do you think of a classmethod to turn an actual file path into a resource path? Ideally with installed packages support so that just passing `__file__` to it would work. And raising `ValueError` if the provided path is not a subpath of the data path.

st3/sublime_lib/pure_resource_path.py
Thom1729: Done.

st3/sublime_lib/pure_resource_path.py
Thom1729: Done now. :-)

st3/sublime_lib/pure_resource_path.py
Thom1729: I fixed the punctuation. Examples would probably be good, but not absolutely required.

st3/sublime_lib/pure_resource_path.py
Thom1729: Fixed.

st3/sublime_lib/path_compat.py
Thom1729: Vendor'd.

st3/sublime_lib/resource_path.py
Thom1729: Even in that case, the path could still point to a file. For example, you could have an installed (zipped) package containing the resource `/Packages/Foo/bar/baz` and an overriding file at `/Packages/Foo/bar`. Then, the resource path `/Packages/Foo/bar` maps to a file but also has descendants. In a filesystem, files and directories are exclusive categories — a path may point to one or the other but never both. Because of this, I think that calling that path a “directory” could lead users to make bad assumptions.

FichteFoll: Ugh, yeah, that would suck.

While we could argue that override packages always override the corresponding installed package, I really don't know how to handle this because most likely ST will *still* see all sub-resources in the installed package where it is a "directory" despite the override being a file. I've never come across this situation, but it is a possibility.

For now, let's concentrate on actionable features and leave this for later.

Thom1729: Would you mind going through and resolving some of the conversations?

FichteFoll: I was hoping the outdated ones would get resolved automatically, but I guess they disabled that once resolving was implemented.

tests/test_pure_resource_path.py
FichteFoll: For clarification:

```
In [3]: pathlib.PureWindowsPath("").root
Out[3]: ''
```

st3/sublime_lib/pure_resource_path.py
FichteFoll: Still not seeing it :confused:

FichteFoll:
Marked all resolved discussions as ... resolved. Do you need a checklist with remaining items to consider or is this enough for you to work with?

st3/sublime_lib/pure_resource_path.py
Thom1729: Added info.

Thom1729:
After some reflection, I think that we should go with the empty string as the global root, with `Packages` and `Cache` as children. The empty path isn't very useful, but I think that it's simpler and less confusing than treating `Packages` and `Cache` as roots.

Then, `PurePath.root` can be removed (because all paths are absolute and have a root of `''`). I think that named properties to grab the first and second components would still be useful. I'm not sure what to call the first, and there are open questions about the semantics of the second. It might be best to just remove those for now; we can always add them back.

Globs should now be feature-complete. (Still need a few more tests.) The `match` method accepts `**` patterns, though pathlib's does not.

I think it does make sense to combine `PureResourcePath` into `ResourcePath`. It's not like we're publishing this on PyPi for non-Sublime users.

> What do you think of a classmethod to turn an actual file path into a resource path? Ideally with installed packages support so that just passing __file__ to it would work. And raising ValueError if the provided path is not a subpath of the data path.

Sounds reasonable. I think we're close enough on this PR that it might make sense to put that feature in a follow-up PR.

It sounds like before merging, we should:

- Nix `PureResourcePath`.
- Add more glob tests.
- Document globs better (perhaps just to say that they should work like pathlib's).
- Remove `root` and `package` (or fix name/semantics).

Please let me know if I've forgotten something.

st3/sublime_lib/resource_path.py
```diff
@@ -40,7 +39,11 @@ def file_path(self):
 
     def exists(self):
         """Return ``True`` if there is a resource at this path, or ``False``
-        otherwise."""
+        otherwise.
+
+        The resource system does not keep track of directories. Even if a path
+        does not contain a resource, there may be resources beneath that path.
```
FichteFoll: `does not represent` or `point to`

st3/sublime_lib/resource_path.py
```diff
@@ -80,11 +83,12 @@ def rglob(self, pattern):
...
         if pattern.startswith('/'):
             raise NotImplementedError("Non-relative patterns are unsupported")
+
         return self.glob('**/' + pattern)
 
     def children(self):
-        """Return a list of ResourcePaths that are direct children of this path
-        and that contain a resource at or beneath that path."""
+        """Return a list of paths that are direct children of this path and
+        contain a resource at or beneath that path."""
```
FichteFoll: `represent` or `point to`

st3/sublime_lib/resource_path.py
```diff
@@ -18,7 +21,148 @@ def glob_resources(pattern):
...
+
+    @property
+    def stem(self):
+        """The final path component without its suffix, or the empty string if
+        the path is empty."""
+        return posixpath.splitext(self.name)[0]
+
+    @property
+    def root(self):
+        """The first path component (usually `Packages` or `Cache`)."""
```
FichteFoll: The values should be in quotes since they are strings (within code blocks).

Also, sphinx uses reST, which requires two backticks for inline code.

FichteFoll:
- +1 for removing `PureResourcePath`.

- I'd be okay with `package` remaining as I find it to be quite useful in several occasions (i.e. I could have used it just yesterday). Imo it should return the second part only if the first is "Packages" and otherwise return `None` or raise an exception. Not sure which I prefer.

- Since the cache and the packages paths are actually quite different and do not "physically" share the same root, I believe it'd be in our interest to have these two be considered the root of the path, i.e. the first part, and also disallow empty resource paths. And all other roots. Following that, the parent of `ResourcePath("Packages")` is itself.

st3/sublime_lib/resource_path.py

st3/sublime_lib/resource_path.py

st3/sublime_lib/resource_path.py

Thom1729:
I've moved `glob_resources` to a staticmethod because `ResourcePath().glob` is no longer available.

> Imo it should return the second part only if the first is "Packages" and otherwise return None or raise an exception. Not sure which I prefer.

I lean toward `None` because it surprises me when properties raise exceptions.

> What do you think of PEP 257?

I haven't seen that, though I can see in it the conventions I've tried to emulate. It might be worth going through and standardizing the docs at some point using that PEP and semantic linefeeds.

I think that's everything we've identified for this issue?

Thom1729:
Should be good. The semantic linefeeds are growing on me.

Also, I figured out how to get the coverage tool to say which lines are missing, so it should be easy to fix up other modules.

PR #53, opened 2018-11-15
Added new_window and close_window.

Thom1729:
By analogy to `view_utils`.

`new_window` creates a new Window and returns it. This is apparently not a core feature. I'm not sure what the failure modes are. deathaxe mentioned on Discord that the Window might not be created immediately if the command is invoked from a non-main thread. If so, then we might to define another function that either a) waits for the new window or b) takes a callback.

`new_window` also takes a pile of convenience arguments (nothing complicated).

`close_window`, like `close_view`, closes a Window without showing a dialog. If there are views with unsaved changes, it will discard changes if `force` is True or raise `ValueError` otherwise.

### DIFF
--- a/st3/sublime_lib/__init__.py
+++ b/st3/sublime_lib/__init__.py
@@ -2,5 +2,4 @@ from .settings_dict import SettingsDict  # noqa: F401
 from .settings_dict import NamedSettingsDict  # noqa: F401
 from .view_stream import ViewStream  # noqa: F401
 from .output_panel import OutputPanel  # noqa: F401
-from .window_utils import new_window  # noqa: F401
-from .window_utils import close_window  # noqa: F401
+from .window_utils import new_window, close_window  # noqa: F401
--- a/st3/sublime_lib/__init__.py
+++ b/st3/sublime_lib/__init__.py
@@ -2,3 +2,5 @@ from .settings_dict import SettingsDict  # noqa: F401
 from .settings_dict import NamedSettingsDict  # noqa: F401
 from .view_stream import ViewStream  # noqa: F401
 from .output_panel import OutputPanel  # noqa: F401
+from .window_utils import new_window  # noqa: F401
+from .window_utils import close_window  # noqa: F401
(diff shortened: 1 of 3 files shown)


st3/sublime_lib/window_utils.py
```diff
@@ -0,0 +1,64 @@
...
+    :raises RuntimeError: if the window is not created for any reason.
+    """
+    original_ids = set(window.id() for window in sublime.windows())
+
+    sublime.run_command('new_window')
+
+    try:
+        window = next(window for window in sublime.windows() if window.id() not in original_ids)
+    except StopIteration:
+        raise RuntimeError("Window not created.")
```
FichteFoll: We'll want `raise … from None` here.

st3/sublime_lib/__init__.py
```diff
@@ -2,3 +2,5 @@
 from .settings_dict import NamedSettingsDict  # noqa: F401
 from .view_stream import ViewStream  # noqa: F401
 from .output_panel import OutputPanel  # noqa: F401
+from .window_utils import new_window  # noqa: F401
+from .window_utils import close_window  # noqa: F401
```
FichteFoll: Why not have these on a single line?

Thom1729: For some reason, I thought that flake would complain. It does not.

st3/sublime_lib/window_utils.py
Thom1729: Today I learned.

Thom1729:
I *think* that `sublime.run_command('new_window')` should be blocking, and I'm not 100% sure about thread-safety. I have yet to test the non-main-thread case.

FichteFoll:
In any case, if we find out it's not working, we can investigate more. The easiest would be to test in a `set_timeout_async` callback. And if that works, within a user-created thread. If any of these don't work, we should document them until we find a proper solution. Or you, rather, as I don't realistically see myself working on this in a reasonable time frame. I'm content with reviewing, though.

I'll squash-merge this anyway. We can adjust/enhanced later.

FichteFoll:
Should probably start sorting imports in `__init__`.

PR #55, opened 2018-11-20
Added show_cursor to ViewStream.

Thom1729:
Allows you to implicitly or explicitly scroll to show the cursor.

In hindsight, I'm not sure whether the explicit `show_cursor` arguments are necessary. If the user wants to show the cursor after calling `write` or somesuch, they can call `show_cursor()`. I'll probably remove those.

### DIFF
--- a/st3/sublime_lib/view_stream.py
+++ b/st3/sublime_lib/view_stream.py
@@ -36,10 +36,9 @@ class ViewStream(TextIOBase):
     will write to the view even if it is read-only. Otherwise, those methods
     will raise :exc:`ValueError`.
 
-    :argument auto_show_cursor: If ``True``, then any method that would move the
+    :argument auto_show_cursor: If ``True``, then any method that moves the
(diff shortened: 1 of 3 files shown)


Thom1729:
Removed `show_cursor` and renamed `auto_show_cursor` to `follow_cursor`.

st3/sublime_lib/view_stream.py
```diff
@@ -117,12 +127,14 @@ def write(self, s):
...
         old_size = self.view.size()
         self.view.run_command('insert', {'characters': s})
+        self._maybe_show_cursor()
         return self.view.size() - old_size
 
     def print(self, *objects, sep=' ', end='\n'):
         """Shorthand for :func:`print()` passing this ViewStream as the `file`
         argument."""
         print(*objects, file=self, sep=sep, end=end)
+        self._maybe_show_cursor()
```
FichteFoll: Should be redundant because `print` calls the `.write` method which already has this call.

PR #60, opened 2018-11-30
Fixed flake error in syntax.

Thom1729:
Not sure why flake8 started complaining about this now.


FichteFoll:
Probably because your version updated and we override the ignore list in `.flake8`.
I (and PEP8) prefer line breaks before the binary operator btw (so ignoring W503).

FichteFoll:
Also, we don't actually have a linter run in your CI config.

PR #63, opened 2018-11-30
Improved flag docs and implemented from_strings.

Thom1729:


Thom1729:
I'm planning to do a final pass over the docs after everything else is done.

FichteFoll:
Yeah, I was thinking the same.

PR #64, opened 2018-11-30
Improve __init__.py.

Thom1729:
- Alphabetize submodules.
- Import public functions from `syntax` and `view_utils`.
- Modify tests to import from `sublime_lib` when possible.

The contents of `flags` and `encoding` are still not exported directly by `sublime_lib`. For `flags`, this probably makes sense. For `encoding`, we'd probably have to export the functions with different names. I am okay with punting on that for now.


FichteFoll:
>The contents of flags and encoding are still not exported directly by sublime_lib.

I agree with your reasoning on this. `flags` should remain in a submodule and `encoding` will probably used very rarely, if at all, so we don't need its members to pollute the module namespace.

PR #66, opened 2018-11-30
Implemented ResourcePath.from_file_path.

Thom1729:
I reworded the docs so that it isn't a breaking change if Sublime adds a new resource path root.

It could conceivably be useful to have a classmethod of ResourcePath that returns a tuple of known roots.


st3/sublime_lib/resource_path.py
```diff
@@ -11,6 +11,12 @@
 __all__ = ['ResourcePath']
 
 
+RESOURCE_ROOTS = {
+    'Packages': sublime.packages_path(),
```
FichteFoll: This is not good. Users will usually import `sublime_lib` when their module is loaded, meaning this is potentially (or likely, even) evaluated before the API has loaded and yields `None`.

st3/sublime_lib/resource_path.py
```diff
@@ -217,23 +246,18 @@ def with_suffix(self, suffix):
...
-        elif self.root == 'Cache':
-            base = sublime.cache_path()
-        else:
-            raise ValueError("%r is not a packages or cache path" % (self,))
-
-        return Path(base).joinpath(*self.parts[1:])
+        try:
+            return Path(RESOURCE_ROOTS[self.root]).joinpath(*self.parts[1:])
+        except KeyError:
+            raise ValueError("Can't find a filesystem path for {!r}.".format(self.root))
```
```suggestion
            raise ValueError("Can't find a filesystem path for {!r}.".format(self.root)) from None
```

(might exceed maximum line lenght?)

st3/sublime_lib/resource_path.py
Thom1729: Fixed.

FichteFoll:
I think we should also add error checking regarding usage of `ResourcePath.from_file_path` before the API has loaded. I'm sure users will be confused about seeing `TypeError: expected str, bytes or os.PathLike object, not NoneType`.

Thom1729:
Is there a good way to check for the API? `plugin_loaded` won't work for a dependency.

FichteFoll:
`sublime.packages_path()` will just return `None`.

Thom1729:
This seems like a cross-cutting concern. There are a number of places where we're calling API methods that might fail if the API is not loaded. If any one of them is called when the API is not loaded, then I think the right thing to do would be to raise an exception. I think that this is something that we should address with a general solution for the entire library (and possibly expose that solution as a feature). On the other hand, I don't think that users are more likely to hit `from_file_path` before the API is ready than they are to hit any other feature that would fail.

If we think it's important to do this for 1.2, then we should probably break it out into a separate issue.

FichteFoll:
Yeah, I forgot we use this in other places as well. Maybe we should just clarify that the library is generally only safe to use when the API has loaded in the docs. Would certainly help reduce boilerplate as checking for API availability every time we use it is bothersome.

Let's wait for reports of people having problems with this first.

PR #68, opened 2018-11-30
Added _util directory.

Thom1729:
~~These utilities should have tests.~~

`glob_util` is probably a good candidate for this directory. Probably also some of the stuff from `settings_dict` and `syntax` (roughly, the stuff that's dragging down the coverage).


st3/sublime_lib/_util/collection_util.py
```diff
@@ -0,0 +1,64 @@
+from collections.abc import Mapping, Sequence
```
FichteFoll: Rename to `collections.py` imo. "util" is already part of the module path.

st3/sublime_lib/_util/collection_util.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+__all__ = ['projection', 'get_selector', 'isiterable', 'ismapping', 'is_sequence_not_str']
+
+
+def projection(d, keys):
+    """
+    Return a new :class:`dict` with keys restricted to values in `keys`.
```
FichteFoll: should mention `d` here, e.g. ```keys of ``d`` ```.

Also ``` ``keys`` ```.

```suggestion
    Return a new :class:`dict` with keys of ``d`` restricted to values in ``keys``.
```

st3/sublime_lib/_util/collection_util.py
```diff
@@ -0,0 +1,64 @@
...
+
+    .. code-block:: python
+
+       >>> projection({'a': 1, 'b': 2}, ['b'])
+       {'b': 2}
+
+    If `keys` is a :class:`dict`, then it maps keys of the original dict to
```
FichteFoll:
```suggestion
    If ``keys`` is a :class:`dict`, then it maps keys of the original dict to
```

st3/sublime_lib/_util/collection_util.py
Thom1729: Done.

st3/sublime_lib/_util/collection_util.py
Thom1729: Done.

st3/sublime_lib/_util/collection_util.py
Thom1729: Done.

Thom1729:
Addressed reviews and moved `glob_util`.

PR #69, opened 2018-12-01
Improve documentation for show_selection_panel, view_stream, and window_utils.

Thom1729:


FichteFoll:
Should we consider running pydocstyle/flake8-doc*whatever on CI? I think that's useful for libraries like this.

Thom1729:
That would probably be a good idea, but I'm not experienced enough with the CI stuff to feel confident about getting it running for 1.2. I'd like to do this for the next release, though.

FichteFoll:
So, uh, what's the status on this? Still WIP? Should I review already?

Thom1729:
This could be merged. There are more docs to improve, but that could be done in another PR.

st3/sublime_lib/view_stream.py
```diff
@@ -165,18 +179,24 @@ def _seek(self, offset):
 
     @guard_validity
     def seek_start(self):
-        """Move the cursor in the view to before the first character."""
```
FichteFoll: I like this short style for single-line docstrings more. But this is mostly your code, so it's just a suggestion. PEP-257 also mandates this, however.

By the way, I also prefer to have the first line of the docstring on the same line as the opening punctuation because I have a plugin that folds all docstrings when I open Python files except for the summary line. It's capable of folding the empty first line up to the actual start of the summary line, but when I want to unfold it using fold arrows, ST will unfold the collapsed whitespace first instead of the remaining body of the docstring because that comes first.
And I think having the body in a separate paragraph is more visually pleasing (applies to the previous change in this file).

Thom1729: This makes sense.

Previously, I've been wrapping comments after column 80. With semantic newlines, most comment lines won't be this long in the first place, so I think it makes sense to wrap at 100 instead (like non-comment lines). This will prevent most summary lines from wrapping, which should be helpful when reading folded code.

PR #70, opened 2018-12-02
Improved prettification of special method docs.

Thom1729:
I'll want to do more hacking on the Sphinx output at some point, but this should be an improvement.


FichteFoll:
Kinda hard to review this from just code and without screenshots. I'm too lazy to pull this branch and build the docs, so I'll just take your word for it.

PR #71, opened 2018-12-02
Sort resource paths by load order.

### LINKED ISSUE #67: Order ResourcePath by load order?
Currently, ResourcePath is ordered lexicographically. This is rather arbitrary. Instead, we could order paths in load order (i.e. Default, everything else, User). This would be immediately useful for, e.g., the `syntax` submodule, and probably more useful to users than lexicographic ordering.

Thom1729:
Closes #67.

A resource's package is significant even if the resource is in `Cache`, so I think it makes sense to expose the `package` property for those paths. Then, I don't see any reason to disable it for other paths.

The `syntax` submodule cares about load order, so I improved the implementation to use `ResourcePath`.


st3/sublime_lib/resource_path.py
```diff
@@ -171,13 +201,12 @@ def root(self):
...
-        The name of the package the path is within,
-        or ``None`` if the path is not beneath ``'Packages'``.
+        The name of the package the path is within.
         """
-        if self.root == 'Packages' and len(self._parts) >= 2:
+        if len(self._parts) >= 2:
             return self._parts[1]
         else:
-            return None
+            return ''
```
FichteFoll: What's the rationale for this change?

If we're about semantics, a `None` more clearly indicates that the path didn't contain a package rather than an empty string. In reality, you'd most likely just `if path.package:` anyway, though.

st3/sublime_lib/syntax.py
```diff
@@ -52,14 +53,14 @@ def get_xml_metadata(text):
...
-        meta = get_xml_metadata(sublime.load_binary_resource(path))
+    if path.suffix == '.sublime-syntax':
+        meta = get_yaml_metadata(path.read_text())
+    elif path.suffix == '.tmLanguage':
+        meta = get_xml_metadata(path.read_bytes())
     else:
         raise TypeError("%s is not a syntax definition." % path)
 
-    return SyntaxInfo(path=path, **meta)
+    return SyntaxInfo(path=str(path), **meta)
```
FichteFoll: Would be useful if we could just use the `ResourcePath` here, but that'd be breaking. :cry:

st3/sublime_lib/resource_path.py
```diff
@@ -82,9 +94,27 @@ def __str__(self):
     def __eq__(self, other):
         return isinstance(other, ResourcePath) and self._parts == other.parts
 
+    def _sort_key(self):
+        return (
+            ROOT_ORDER.get(self.root, 0),
+            self.root,
+            PACKAGE_ORDER.get(self.package, 0),
+            self._parts[1:]
```
FichteFoll: Something we haven't considered yet: Resource paths are sorted files-first, i.e. all files come before any subdirectories.

```py
>>> sublime.find_resources("test.test")
['Packages/User/Test/test.test', 'Packages/User/Test/A/test.test']
```

And it's case-insensitive:
```py
>>> sublime.find_resources("test.test")
['Packages/User/Test/test.test', 'Packages/User/Test/A/test.test', 'Packages/User/Test/a/test.test', 'Packages/User/Test/B/test.test']
```

Also, this should start at `[2:]`.

.flake8
```diff
@@ -1,6 +1,6 @@
 [flake8]
 
-ignore = E731
+ignore = E731, W503
```
FichteFoll: Could you add comments for the meaning of these codes? Coincidentally I know what W503 is, but I don't know E731 by heart and having to look these up is quite annoying.

st3/sublime_lib/syntax.py
Thom1729: Same.

.flake8
Thom1729: Done.

st3/sublime_lib/resource_path.py
Thom1729: Agreed

st3/sublime_lib/resource_path.py
Thom1729: It looks like it's even more complicated than that. Physical files in `Packages` seem to be sorted after resources in `Installed Packages`, and I have a sneaking suspicion that resources in zipfiles are loaded in the zipfile directory order. I will investigate this probably tomorrow.

st3/sublime_lib/resource_path.py
FichteFoll: Please also add this to the docstring.

Thom1729: Fixed.

Thom1729:
Given:

1. Load order seems to be the natural ordering of resource paths.
2. Load order is complicated and possibly not fully understood.
3. Changing the ordering relation would be a breaking change.
4. Using `ResourcePath` in `syntax` is a good thing either way.

I suggest that for the time being we:

- Pull `__lt__` and make `ResourcePath` not orderable.
- Preserve the order of `sublime.find_resources` when returning a list of `ResourcePaths` rather than explicitly sorting.
- For `children`, explicitly document that the order is not guaranteed (because not all children necessarily point to extant resources).

Then, we can add ordering (in 1.2 or another release) when we work out the details and if it won't be a mess. Meanwhile, if a user wants to sort a list of `ResourcePath`s, they can specify their own sort key (such as `str()`).

Sound reasonable?

FichteFoll:
Yes to all.

We could still implement a method to order resources properly (using cached (installed/default) packages lists), but that should be somewhere else and not in `__lt__`.

Thom1729:
Removed explicit comparison and added a function for use in `syntax`.

Because `sublime.load_resources()` apparently has no order guarantees, I didn't add anything about ordering to the docs, one way or the other.

Given that we're hopefully expecting a new API method to solve the problem for syntaxes, this should probably do the trick for now.

st3/sublime_lib/resource_path.py
```diff
@@ -11,7 +10,35 @@
...
+        elif package == 'User':
+            package_order = 1
+        elif package in installed_packages:
+            package_order = -1
+        elif package in default_packages:
+            package_order = -2
+        else:
+            package_order = 0
+
+        return (ROOT_ORDER[path.root], path.root, package_order) + path.parts[1:]
```
```suggestion
        return (ROOT_ORDER[path.root], package_order) + path.parts[1:]
```
Like this, right?

st3/sublime_lib/resource_path.py
```diff
@@ -11,7 +10,35 @@
...
 
 
-@total_ordering
+ROOT_ORDER = {
+    'Packages': -2,
+    'Cache': -1,
+}
+
+
+def sort_by_load_order(paths):
```
FichteFoll: Would like to have a docstring here explaining roughly what it does (i.e. to what point it mirrors the internal order).

Case-insensitivity is missing, as far as I can tell.

st3/sublime_lib/resource_path.py
Thom1729: That wouldn't sort nonstandard roots (e.g. `A/Foo` before `Z/Foo`). I'm not sure whether the order really matters, but even for the current limited use I think it's best to do a total sort.

Thom1729: Right, duh. Fixed.

st3/sublime_lib/resource_path.py
Thom1729: After more testing, it looks like the current implementation still sorts some paths differently from `sublime.find_resources` (e.g. `.DS_Store`). If `sublime.find_resources` is the closest we have to a canonical load order, then I may as well just modify `syntax` to use that ordering directly and remove all of the explicit sorting code. I wish I could be sure that that's absolutely correct, but it might be good enough to tide us over until the new API method obviates the whole thing.

Thom1729:
Latest commit removes all load order sorting and modifies `syntax` to use the order of `sublime.load_resources`.

PR #72, opened 2018-12-02
Fix bugs in `syntax` and improve tests.

### LINKED ISSUE #73: Ignore shadowed `tmLanguage` files.
When there are a `sublime-syntax` and `tmLanguage` file with otherwise-identical paths, the `sublime-syntax` will “shadow” the `tmLanguage` for syntax-related purposes:

- It won't appear in the command palette.
- It won't appear in the syntax menu.
- It won't be found when including syntaxes by scope.
- It won't be found when picking a syntax by file extension.

This is true regardless of the contents of the two files (even if one is hidden, or invalid).

(In addition, if you manually set `view.settings().set('syntax', 'my.tmLanguage')`, the highlighter will use the `sublime-syntax`. This probably doesn't affect us.)

Currently, neither `list_syntaxes` nor `get_syntax_for_scope` acknowledges this. Both will return shadowed `tmLanguage` definitions. I believe that this behavior is incorrect (i.e. that changing it is a bug fix). Fixing it is probably easy on top of #71 and #72.

`get_syntax_for_scope` definitely should ignore shadowed `tmLanguage` files. `list_syntaxes` probably should as well. We could conceivably add a flag to `list_syntaxes` to include shadowed syntaxes, but I'm not sure that it would be useful.

### LINKED ISSUE #76: Handle `hidden-tmLanguage`
Questions:

- Is the shadowing behavior the same as `tmLanguage`?
- Does the `hidden` key do anything?

Thom1729:
Fix #73. Fix #76.


Thom1729:
Fix #73. Fix #76.

Also handle the following quirks:

- If a `sublime-syntax` or `tmLanguage` file has no `name` key, or the `name` key is null or the empty string, use the filename stem as the syntax name. (Note: Sublime will complain in an error dialog if a `tmLanguage` lacks a `name` key altogether, so I didn't add this case to the unit tests.)
- `hidden-tmLanguage` syntaxes always use the filename stem as the syntax name, regardless of the `name` key.
- `hidden-tmLanguage` syntaxes are always hidden, regardless of the `hidden` key.

Also modified `ResourcePath.read_text|bytes` to include the path string when raising `FileNotFoundError`.

st3/sublime_lib/_util/yaml.py
```diff
@@ -0,0 +1,29 @@
...
+import re
+
+
+__all__ = ['parse_simple_top_level_keys']
+
+
+def parse_simple_top_level_keys(text):
+    return dict(
+        map(_parse_yaml_value, match.groups())
+        for match in re.finditer(r'(?m)^(\S.*?\s*): *(.+) *$', text)
```
FichteFoll: Not changed in this PR, but we might want to check for comments, although I doubt people are using them in this line

st3/sublime_lib/syntax.py
```diff
@@ -1,10 +1,7 @@
-import sublime
-
-import re
 from collections import namedtuple
 import plistlib
 
-from ._util.collections import projection
+from ._util.yaml import parse_simple_top_level_keys
```
FichteFoll: Maybe call it `simple_yaml`?

st3/sublime_lib/syntax.py
```diff
@@ -15,57 +12,58 @@
...
+        path=str(path),
+        **SYNTAX_TYPES[path.suffix](path)
+    )
+
 
-    return SyntaxInfo(path=str(path), **meta)
+def shadowed(path):
+    return (
+        path.suffix in {'.tmLanguage', '.hidden-tmLanguage'}
+        and path.with_suffix('.sublime-syntax').exists()
```
FichteFoll: We could save this call (that ends up making another `find_resources` call for every non-sublime-syntax syntax) by caching the initial `ResourcePath.glob_resources('')` call in line 85 and checking whether that contains the sublime-syntax-suffixed path.

I'd probably inine this function into `list_syntaxes` after `glob_resources` was assigned to a variable.

st3/sublime_lib/syntax.py
Thom1729: Done.

st3/sublime_lib/_util/yaml.py
Thom1729: Done.

st3/sublime_lib/syntax.py
Thom1729: Done.

Issue #74, opened 2018-12-05
Generate also a single API documentation page like Sublime Text API page


evandrocoan:
Currently you need to browse through all submodules pages:
![image](https://user-images.githubusercontent.com/5332158/49539872-c7397280-f8b5-11e8-8e1b-059b097259d2.png)

But, It would be nice if I can access a single page with everything, like the Sublime API page: https://www.sublimetext.com/docs/3/api_reference.html

FichteFoll:
Adding this here since it's most relevant:

Many functions, classes and methods provided require the API to have loaded. Thus, it makes sense to mention this on the front page instead of every exported function or class.


Furthermore, we export most of the things that are stored in submodules from the main module and those locations should be the ones documented.

FichteFoll:
Done via #77.

FichteFoll:
Well, except for this, I guess?

>Many functions, classes and methods provided require the API to have loaded. Thus, it makes sense to mention this on the front page instead of every exported function or class.

PR #77, opened 2018-12-11
Reformat docs.

Thom1729:
For #74.

Docs are now on a single page.

I totally redid the theme. The “Read the Docs” theme was very noisy. We can add more ornamentation as desired.

Notably, I removed the search function. With all the docs on one page, the browser's built-in search is good enough.

Feedback requested on organization and aesthetics.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,7 @@ MODULEDIR     = st3
 .PHONY: source clean
 
 source:
-	SPHINX_APIDOC_OPTIONS=members,no-special-members sphinx-apidoc --separate --force --no-toc -o "$(SOURCEDIR)/modules" "$(MODULEDIR)"
+	SPHINX_APIDOC_OPTIONS=members sphinx-apidoc --force --module-first -o "$(SOURCEDIR)/modules" "$(MODULEDIR)"
 
 html:
 	sphinx-build -M html "$(SOURCEDIR)" "$(BUILDDIR)"
--- a/docs/source/_static/style.css
+++ b/docs/source/_static/style.css
@@ -1,27 +1,62 @@
-code, dt {
-    font-size: inherit !important;
+.related { display: none; }
+
+html {
+    font: 20px Roboto;
+    margin: 0;
 }
 
-ul.simple {
-    /*columns: 10em;*/
+body {
+    margin: 0 auto;
+    max-width: 33rem;
+
+    font-size: 15px;
+    line-height: 1rem;
 }
 
-ul.simple > li > :last-child {
-    margin-bottom: 0 !important;
+code, pre {
+    font-family: "Roboto Mono";
 }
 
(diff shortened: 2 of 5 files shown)


Thom1729:
@evandrocoan Thoughts?

FichteFoll:
Liking the new simple layout, but I'm missing some kind of introduction paragraph before we're going straight at the module docs. At least something like we have in the readme, imo.

evandrocoan:
> FichteFoll: Liking the new simple layout and toc, but I'm missing some kind of introduction paragraph before we're going straight at the module docs. At least something like we have in the readme, imo.

The Sphinx builder could copy the README.md, or at least some delimited part of it into the generated html.


> Thom1729: @evandrocoan Thoughts?

I never worked with Sphinx so far, then, I missed something like a `Makefile` on the `docs` directory:
image

After searching for Sphinx, I found this http://www.sphinx-doc.org/en/1.5/tutorial.html and used this command `sphinx-build -b html source builddir`

This is the quick make file I came up with:
```Makefile
## Usage:
##   make <target> [options]
##
## Targets:
##   all               generate the documentation on the builddir directory
##   help              show this usage help
##
## Options:
##   -p                show all commands run by make, including all variables
##   -n                similar to --dry-run, it will only display the commands,
##                     but not run them.
##
## For more explanations see:
##   http://www.oreilly.com/openbook/make3/book/ch12.pdf
all:
	sphinx-build -b html source builddir

# Print the usage instructions
# https://gist.github.com/prwhite/8168133
help:
	@fgrep -h "##" $(MAKEFILE_LIST) | fgrep -v fgrep | sed -e 's/\\$$//' | sed -e 's/##//'
```

If I did it right, this was the result:
image

I did not had though about adding a sidebar table of contents. Nice idea. Optionally, I think it would be nice if the sidebar had a indicator, pointing to where I am on the page while I am "scrolling" around. For example:
fsg

I found this example on:
1. https://stackoverflow.com/questions/27610892/highlight-menu-on-scrolling-if-reach-div
1. http://jsfiddle.net/vu6hN/28/

Thom1729:
There's a Makefile in the package root — just run `make html`. Your build looks right, though.

I like the scrolling idea. I think I might try to remove more of the original theming before adding new functionality. This could be done in a patch release, I assume.

evandrocoan:
I just found another example for the table of contents: 
1. https://apps.ankiweb.net/docs/changes.html#changes-in-2.1.0
1. fsg

FichteFoll:
As we develop more and more, we might need to start thinking about versioning the docs on `gh-pages` or only updating them when a tag is pushed. Currently, it will always build and render docs from `master`.

Feel free to create an issue for this if you consider this to be worthwhile, @Thom1729.

For the TOC-suggestion, see #84.

PR #79, opened 2018-12-12
Added line_endings argument to new_view, etc.

Thom1729:
Valid values are `'Unix'`, `'Windows'`, and `'CR'`. Following Sublime, we don't verify the case (Sublime will normalize the case internally).

### DIFF
--- a/st3/sublime_lib/view_utils.py
+++ b/st3/sublime_lib/view_utils.py
@@ -1,10 +1,49 @@
 import inspect
 
+from .vendor.python.enum import Enum
 from .syntax import get_syntax_for_scope
 from .encodings import to_sublime
 
 
-__all__ = ['new_view', 'close_view']
+__all__ = ['LineEnding', 'new_view', 'close_view']
+
+
+class LineEnding(Enum):
+    """An :class:`enum` of line endings supported by Sublime Text.
+
+    The :class:`LineEnding` constructor accepts either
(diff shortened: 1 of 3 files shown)


st3/sublime_lib/view_utils.py
```diff
@@ -114,9 +126,14 @@ def set_view_options(
...
+        view.set_line_endings(line_endings)
+
 
 VIEW_OPTIONS = {
     name
     for name, param in inspect.signature(set_view_options).parameters.items()
     if param.kind == inspect.Parameter.KEYWORD_ONLY
 }
+
+LINE_ENDINGS = {'unix', 'windows', 'cr'}
```
FichteFoll: This should be defined at the top, imo.

st3/sublime_lib/view_utils.py
```diff
@@ -17,6 +17,9 @@ def new_view(window, **kwargs):
 
     :argument encoding: The encoding that the view should use when saving.
 
+    :argument line_endings: The kind of line endings to use:
+        ``'Unix'``, ``'Windows'``, or ``'CR'``.
```
FichteFoll: I wonder whether using the respective string representation would be clearer. I mean, the underlying API uses these values, but everything within Python just uses `"\r\n"` or `"\n"` and I like that a lot more.

Thom1729:
What about an `Enum` for line endings? We'd convert `'unix'` (case-insensitively) or `'\n'` into `LineEnding.Unix` into `'Unix'`.

FichteFoll:
Fancy.

PR #88, opened 2018-12-16
Fixed makefile for new docs.

### LINKED ISSUE #85: Not put the Makefile on the root of the project
Just some thoughts mine about it.

I would not put the `Makefile` on the root of the project because it is not a place I would look for it if I would like to build the `docs`, because:
1. This is a python project, so most of times does not make sense to have a `Makefile`. Then, while looking for how to build the `docs`, I did even thought about looking for it on the root of the project (https://github.com/SublimeText/sublime_lib/pull/77 Reformat docs). 
1. The only thing the `Makefile` builds is the docs, then, it would be better to put a  `docs` related file into the `docs` directory. On bigger projects, we could have one `Makefile` per "subproject". If that was the case, our main `Makefile` on the root of the project, would recursively call other make files.
1. The `docs` directory is kinda empty (only the `source` directory inside it), while the root of the project has much more other things, including the `Makefile` which is used exclusively for the `docs` directory, so it should be put inside it for "organization", as in, keep related things together.

### LINKED ISSUE #86: Some Ghosts on the new TOC
image

Thom1729:
Since we have the manual `index.rst`, we don't need `make source`. Should fix #86. Also fixes #85.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,9 +2,9 @@ language: python
 python: 3.6
 
 install:
-  - pip install sphinx sphinx_rtd_theme
+  - pip install sphinx
 
-script: make source html
+script: make clean html
 
 git:
   depth: 1
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ python: 3.6
 install:
   - pip install sphinx
 
-script: make clean html
+script: make -C docs clean html
 
 git:
   depth: 1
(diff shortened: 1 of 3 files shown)


Thom1729:
I've split it into two commits. The GitHub diff tool is confused, but `git log --follow -- docs/Makefile` works.

evandrocoan:
> Thom1729: The GitHub diff tool is confused

It is indeed a bug actually. If failed to create a resume diff including the file renaming. But when individually looking the commits, we can see it showing the file renaming correctly:
image

I am going to report this to them.

PR #89, opened 2018-12-16
Last-minute docs fixes.

Thom1729:
- Added missing `versionadded` and `versionchanged` directives.
- Added `LineEnding` enum to docs.
- Fixed a typo in `NamedSettingsDict`.

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -27,6 +27,7 @@ View utilities
 
 .. autofunction:: sublime_lib.new_view
 .. autofunction:: sublime_lib.close_view
+.. autoclass:: sublime_lib.LineEnding
 
 Window utilities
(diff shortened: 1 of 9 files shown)


Thom1729:
`NOT_GIVEN`, as used in `show_selection_panel` and `SettingsDict.get`, doesn't display well in the docs. I'm replacing it with an explicitly named value that will.

Thom1729:
Use a `NamedValue` `SettingsDict.NO_DEFAULT` as the default value of the `default` argument. This is morally the same as what we were doing before, but better-documented. There shouldn't be any way that this could break anything.

Ideally, we'd exactly replicate the `dict.pop` behavior, but that's not possible in Python. This should be close enough for any practical purpose.

Issue #90, opened 2018-12-17
Lightweight async mini-framework


Thom1729:
The builtin options for deferred actions in Sublime plugins are:

- `set_timeout()` (which can block the main thread).
- `set_timeout_async()` (which can block the single, shared async thread).
- DIY manually using threads.

The relative ease and convenience of the former two options leads devs to use them when a more robust solution would be appropriate. `set_timeout_async()`, in particular, seems to be misunderstood — blocking operations (such as IO or sleep) in a `set_timeout_async()` callback will block everyone else's callbacks.

A simple-but-robust mechanism for deferred and async code could be very helpful to package developers and valuable to end users.

### Requirements

Package developers write deferred or async code to:

- Schedule code to run later.
- Keep blocking operations off of shared threads.
- Yield control to other tasks during expensive operations.
- Asynchronously wait for some condition.
- Interact with external processes such as Node.

Depending on the circumstances, the same task might be run with very different execution strategies. For instance, an expensive task like compiling macros might typically run on its own thread, but to calculate test coverage it must run on the main thread. The developer must be able to switch out execution strategies on the fly.

Some of Sublime's API methods (such as `Window.show_quick_panel`) run deferred callbacks. A developer should be able to suspend their code while awaiting a callback and then continue execution in the same manner as before (e.g. on the same thread).

In order to provide an attractive and convenient alternative to blindly calling `set_timeout_async()`, the API must be extremely simple.

### Non-requirements

Python has a global interpreter lock that prevents more than one thread from running Python code simultaneously. The workaround is multiprocessing. A convenient abstraction for this is outside the scope of this proposal.

Managing shared state in async code can be hard. While we shouldn't introduce unnecessary footguns, nor should we try to solve all of the ones that already exist.

Python threads are cooperative; a single thread can hog the CPU. We can't stop this. We can, however, provide abstractions that avoid unexpected conflicts (e.g. blocking IO in the shared async thread).

### Prior art

Python's [concurrent.futures](https://docs.python.org/3/library/concurrent.futures.html) module provides `Future` and `Executor` abstractions. These are useful primitives, particularly `Future`. In principle, custom `Executor` classes could encapsulate `set_timeout` and `set_timeout_async`. However, the API is low-level. `Future`s are atomic; there is no abstraction for a task containing several units of work. In addition, there is no notion of scheduling or timing.

[...]

PR #93, opened 2019-01-03
[WIP] Added async framework.

Thom1729:
This is a first pass at implementing the ideas outlined in #90.

The core function is `run(function, scheduler)`. `scheduler` may be `set_timeout`, `set_timeout_async`, or any function with the same signature. Two new such schedulers are provided. `thread_scheduler()` starts a new thread, runs the task, and ends the thread. `SharedThreadRunner` maintains a shared thread in which parts of tasks are enqueued.

I think that the core should be fairly solid. The API is a work in progress.

Example tasks:

```python
def delay_task():
    for i in range(5):
        yield 500 # Wait 500 ms
        print(i)

from sublime_lib import show_selection_panel
def selection_task():
    try:
        result = yield promise(lambda future:
            show_selection_panel(
                sublime.active_window(),
                ['Alice', 'Bob', 'Carol'],
                on_select=future.set_result,
                on_cancel=lambda: future.set_exception(ValueError("Cancelled"))
            )
        )
    except ValueError as ex:
        print(ex)
    else:
        print(result)
```

Invocations:

```python
run(delay_task, sublime.set_timeout)
run(delay_task, thread_scheduler())

shared = SharedThreadRunner()
run(delay_task, shared.schedule)
```

### DIFF
--- a/st3/sublime_lib/async.py
+++ b/st3/sublime_lib/async.py
@@ -1,0 +1,163 @@
+from concurrent.futures import Future
+from functools import partial
+from traceback import print_exception
+from inspect import isgenerator
+
+
+def wait_for_condition(condition, period=100, iterations=5):
+    for i in range(iterations):
+        result = condition()
+        if result:
+            return result
+        else:
+            yield period
+
+    raise TimeoutError("Timed out after {}ms.".format(iterations * period))
+
+
+def promise(function):
+    future = Future()
+    function(future)
+    return future
+
+
+def run(function, scheduler):
+    future = Task(function, scheduler).future
+    future.add_done_callback(log_error)
+    future.add_done_callback(lambda future: scheduler(None, 0))
+    return future
+
+
+def log_error(future):
+    ex = future.exception(0)
+    if ex:
+        print_exception(type(ex), ex, ex.__traceback__)
+
+
+class Task():
+    def __init__(self, function, scheduler):
+        self._schedule = scheduler
+        self.future = Future()
+
+        self._schedule(partial(self._start, function), 0)
+
+    def _start(self, function):
+        try:
+            result = function()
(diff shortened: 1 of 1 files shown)


st3/sublime_lib/async.py
```diff
@@ -0,0 +1,163 @@
...
+    def decorator(function):
+        return lambda *args, **kwargs: run(partial(function, *args, **kwargs))
+
+    return decorator
+
+
+    from functools import partial
+    from threading import Thread, Event
+    from time import sleep, monotonic
+    from queue import Queue, PriorityQueue, Empty
```
FichteFoll: What's this block?

st3/sublime_lib/async.py
```diff
@@ -0,0 +1,163 @@
...
+                action.add_done_callback(
+                    lambda f: self._schedule(partial(self._next, f), 0)
+                )
+            else:
+                raise TypeError("Task returned invalid action.")
+
+
+def async_decorator(run):
+    def decorator(function):
+        return lambda *args, **kwargs: run(partial(function, *args, **kwargs))
```
FichteFoll: I guess this should be a proper function wrapped with `functools.wraps`.

st3/sublime_lib/async.py
Thom1729: In the wrong place. Fixed.

st3/sublime_lib/async.py
Thom1729: Fixed.

Thom1729:
Added [WIP] label to clarify that the API is not great and I don't think it's quite ready for release.

Thom1729:
Since Python 3.8 provides `async`, this probably isn't necessary.

PR #96, opened 2019-01-04
Added Panel class.

Thom1729:
The existing `OutputPanel` class contains a lot of functionality that could be useful for any panel, such as the console. This PR factors out that functionality into a new `Panel` class. `OutputPanel` then extends both `ViewStream` and `Panel`. (Thankfully, this isn't a mess in Python as it would be in many other languages.)

The behavior of `OutputPanel` is unchanged, except that when you call inherited `Panel` methods on an `OutputPanel` whose view is invalid, the exception message will be slightly different. (It will still raise the same exception under the same circumstances.)

"Before" example, based on real code with irrelevant bits removed:

```python
if settings["open_console"] and self.window.active_panel() != "console":
    self.window.run_command("show_panel", {"panel": "console"})

try:
    reload_package(pkg_name)
except Exception:
    if settings["open_console_on_failure"]:
        self.window.run_command("show_panel", {"panel": "console"})
    raise
finally:

if settings["close_console_on_success"]:
    self.window.run_command("hide_panel", {"panel": "console"})
```

After:

```python
console = Panel("console")

if settings["open_console"] and not console.is_visible():
    console.show()

try:
    reload_package(pkg_name)
except Exception:
    if settings["open_console_on_failure"]:
        console.show()
    raise
finally:

if settings["close_console_on_success"]:
    console.hide()
```

Notes:
- I factored out `define_guard` under `_util`.
- The OutputPanel tests are functionally the same, but I modified them to use `assertRaises` with a `with` statement, which is much nicer.
- `OutputPanel.full_name` is now redundant with `Panel.panel_name` (but remains, of course, for compatibility).

### DIFF
--- a/st3/sublime_lib/__init__.py
+++ b/st3/sublime_lib/__init__.py
@@ -1,4 +1,4 @@
-from .output_panel import Panel, OutputPanel  # noqa: F401
+from .panel import Panel, OutputPanel  # noqa: F401
 from .resource_path import ResourcePath  # noqa: F401
 from .show_selection_panel import show_selection_panel, NO_SELECTION  # noqa: F401
 from .settings_dict import SettingsDict, NamedSettingsDict  # noqa: F401
--- a/st3/sublime_lib/__init__.py
+++ b/st3/sublime_lib/__init__.py
@@ -1,4 +1,4 @@
-from .output_panel import OutputPanel  # noqa: F401
+from .output_panel import Panel, OutputPanel  # noqa: F401
 from .resource_path import ResourcePath  # noqa: F401
 from .show_selection_panel import show_selection_panel, NO_SELECTION  # noqa: F401
 from .settings_dict import SettingsDict, NamedSettingsDict  # noqa: F401
--- a/st3/sublime_lib/output_panel.py
+++ b/st3/sublime_lib/output_panel.py
@@ -1,12 +1,70 @@
 from .view_stream import ViewStream
 from .view_utils import set_view_options, validate_view_options
+from ._util.guard import define_guard
+
+__all__ = ['Panel', 'OutputPanel']
 
 
-__all__ = ['OutputPanel']
+class Panel():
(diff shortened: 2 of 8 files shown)


st3/sublime_lib/__init__.py
```diff
@@ -1,4 +1,4 @@
-from .output_panel import OutputPanel  # noqa: F401
+from .output_panel import Panel, OutputPanel  # noqa: F401
```
FichteFoll: Should probably rename this module to just `panel`, i.e. the more general name.

FichteFoll:
Btw, you could have waited for a few hours with the earlier release so we could have had these few changes in there as well.

FichteFoll:
Oh, and if we make something redudant, we should deprecate it immediately, imo, and refer to the replacement.

st3/sublime_lib/output_panel.py
```diff
@@ -1,12 +1,70 @@
...
+    :raise ValueError: if `window` has no panel called `panel_name`.
+
+    All :class:`Panel` methods except for :meth:`exists()`
+    will raise a ``ValueError`` if the panel does not exist.
+    Descendant classes may override :meth:`exists()` to customize this behavior.
+    """
+
+    def __init__(self, window, panel_name):
+        self.window = window
+        self.panel_name = panel_name
```
FichteFoll: This attribute should definitely be documented.

Thom1729:
This is new functionality, so I think it might be preferable to hold off and put it in the next minor release rather than in a patch release. Likewise with #95.

st3/sublime_lib/output_panel.py
Thom1729: Done.

st3/sublime_lib/__init__.py
Thom1729: Done.

FichteFoll:
Hm, maybe. I don't have a strong opinion on this. You decide.

PR #98, opened 2019-01-11
Add ResourcePath.copy

Thom1729:
Modeled after `pathlib.Path.rename`.


st3/sublime_lib/resource_path.py
```diff
@@ -359,3 +359,23 @@ def children(self):
                 for resource in self.glob('**')
             )
         ]
+
+    def copy(self, target, exist_ok=True):
+        """
+        Copy this resource to the given `target`.
+
+        If `target` exists and is a file,
```
FichteFoll: Should probably mention that target should be a string or `pathlib.Path` (or, for the future, a [path-like object](https://docs.python.org/3/library/os.html#os.PathLike))?

PR #111, opened 2019-02-16
Implemented missing flags from sublime_plugin.

Thom1729:
It might also be handy to provide operator functionality for `QueryContextOperator`. Not sure of the best API for that.

EDIT: Maybe just `__call__`? E.g., `QueryContextOperator.EQUAL(1, 1)`.

### DIFF
--- a/docs/source/_static/style.css
+++ b/docs/source/_static/style.css
@@ -117,6 +117,7 @@ pre {
     border: 1px solid #33333333;
     border-radius: 2px;
     padding: 0.25rem;
+    font-size: 13px;
 }
 
 div.footer {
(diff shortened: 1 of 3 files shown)


FichteFoll:
For use cases, I could imagine both a classmethod on the enum or an instance method, but `__call__` has some anti-patterns imo when you're coming from the typical context operator parameters (i.e. it looks confusing). You're not going to type the enums out to call a function on them. You're either going to get the enum value though `__call__` of the metaclass or compare the value explicitly (`operator == QueryContextOperator.EQUAL`).

Input variables are `operand` and `operator`, `myval` is what you compare with:
```py
# __call__
QueryContextOperator(operator)(operand, myval)
# classmethod
QueryContextOperator.apply(operator, operand, myval)
# method
QueryContextOperator(operator).apply(operand, myval)
```

I think I prefer the classmethod.

Thom1729:
Implemented as instance method.

PR #112, opened 2019-02-18
Better handling of resource roots.

Thom1729:
Previously, we handled the mapping between file and resource paths using `get_resource_roots()` and `get_installed_resource_roots()`. This PR consolidates the mapping logic into `ResourceRoot` objects, each representing a directory containing either zipped or unzipped packages. `file_path()` and `from_file_path()` merely iterate over the `ResourceRoot` objects and return the first result.

Other tweaks:

- Cache the list of roots instead of recomputing it each time.
- Instead of `pathlib.Path.relative_to`, use a custom implementation that doesn't spend most of its CPU time formatting an error message that we don't care about.
- Use a `wrap_path` function to avoid calling the (relatively expensive) `Path` constructor on `Path` objects.
- Avoid unnecessary reparsing when calling the `ResourcePath` constructor with a `ResourcePath` object as the first argument.

These changes improved the performance of a function that calls `from_file_path()` on all of the paths in `sys.modules` by more than a factor of three.

Currently, the `ResourceRoot` list isn't exposed or documented. This may be a useful primitive for packages that care where resources are located.

### DIFF
--- a/st3/sublime_lib/resource_path.py
+++ b/st3/sublime_lib/resource_path.py
@@ -3,6 +3,7 @@ import sublime
 import posixpath
 from collections import OrderedDict
 import os
+from abc import ABCMeta, abstractmethod
 
 from .vendor.pathlib.pathlib import Path
 from ._util.glob import get_glob_matcher
@@ -16,6 +17,7 @@ def _abs_parts(path):
         return (path.drive, path.root) + path.parts[1:]
     else:
         return path.parts
+
 
 def _file_relative_to(path, base):
     """
@@ -39,17 +41,26 @@ def _file_relative_to(path, base):
         compare = cf(child_parts[:n])
 
     if compare != cf(base_parts):
-        raise ValueError()
+        return None
 
     return child_parts[n:]
 
 
-class ResourceRoot():
+class ResourceRoot(metaclass=ABCMeta):
+    """
+    Represents a directory containing packages.
+    """
     def __init__(self, root, path):
         self.resource_root = ResourcePath(root)
         self.file_root = Path(path)
 
     def resource_to_file_path(self, resource_path):
+        """
+        Given a :class:`ResourcePath`,
+        return the corresponding :class:`Path` within this resource root.
+
+        :raise ValueError: if the :class:`ResourcePath` is not within this resource root.
+        """
         resource_path = ResourcePath(resource_path)
 
         parts = resource_path.relative_to(self.resource_root)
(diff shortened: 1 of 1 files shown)


st3/sublime_lib/resource_path.py
```diff
@@ -11,18 +11,102 @@
...
+    n = len(base_parts)
+    cf = path._flavour.casefold_parts
+
+    if n == 0:
+        compare = (path.root or path.drive)
+    else:
+        compare = cf(child_parts[:n])
+
+    if compare != cf(base_parts):
+        raise ValueError()
```
FichteFoll: The error message being empty has a weird effect on reading.

The first thing you think about is that we might as well return `None` and skip exception handling overhead if we don't care about the error. However, it's not caught in `ResourceRoot.file_to_resource_path` and instead caught in `ResourcePath.from_file_path` and ignored to continue the loop.

From how this pattern is used, I would suggest not raising an exception that nobody will care about and just return `None` and adjusting code paths to care for that (notably `ResourceRoot.file_to_resource_path` and its callee). Should increase performance even more.

st3/sublime_lib/resource_path.py
```diff
@@ -11,18 +11,102 @@
...
+        if parts == ():
+            return self.file_root
+        else:
+            return self._package_file_path(*parts)
+
+    def file_to_resource_path(self, file_path):
+        file_path = wrap_path(file_path)
+
+        if not file_path.is_absolute():
+            raise ValueError("Cannot convert a relative file path to a resource path.")
```
FichteFoll: This should then also return `None`.

st3/sublime_lib/resource_path.py
```diff
@@ -11,18 +11,102 @@
...
+
+        parts = _file_relative_to(file_path, self.file_root)
+        # parts = file_path.relative_to(self.file_root).parts
+        if parts == ():
```
FichteFoll: And this check for `None` and return that.

st3/sublime_lib/resource_path.py
```diff
@@ -11,18 +11,102 @@
...
+    else:
+        compare = cf(child_parts[:n])
+
+    if compare != cf(base_parts):
+        raise ValueError()
+
+    return child_parts[n:]
+
+
+class ResourceRoot():
```
FichteFoll: Some internal documentation for how this is used and what abstract methods it expects would be useful.

st3/sublime_lib/resource_path.py

st3/sublime_lib/resource_path.py
Thom1729: It's fine for this to raise because we can't handle that error anyway.

st3/sublime_lib/resource_path.py

st3/sublime_lib/resource_path.py
```diff
@@ -97,46 +224,43 @@ def from_file_path(cls, file_path):
...
+                for path in (
+                    root.file_to_resource_path(file_path)
+                    for root in get_roots()
+                )
+                if path is not None
+            )
+        except StopIteration:
+            raise ValueError(
+                "Path {!r} does not correspond to any resource path.".format(file_path)
+            )
```
```suggestion
            ) from None
```

I don't like this section. One one hand, it's ingenious (seemingly), but on the other hand it is error-prone (like I showed) and it's hard to get the order of execution right in your head. The double comprehension is boilerplatey.

```py
candidates = root.file_to_resource_path(file_path) for root in get_roots()
path = next(filter(None, candidates), None)
if path:
    return path
else:
    raise ValueError(
        "Path {!r} does not correspond to any resource path.".format(file_path)
    )
```
is easier to read, has less lines and is more explicit.

PR #116, opened 2019-03-01
Add comprehensive type checks

Thom1729:
Includes a stub for the `typing` module. I tried vendoring the module, but it doesn't play well with 3.3, and we don't actually need it to do anything anyway.

Sphinx automatically includes type annotations in the docs. Sometimes the formatting is a mess. This should be addressed.

### DIFF
--- a/st3/sublime_lib/_util/collections.py
+++ b/st3/sublime_lib/_util/collections.py
@@ -16,7 +16,10 @@ except ImportError:
 __all__ = ['projection', 'get_selector', 'isiterable', 'ismapping', 'is_sequence_not_str']
 
 
-def projection(d: 'Dict[str, _V]', keys: 'Union[Dict[str, str], Iterable[str]]') -> 'Dict[str, _V]':
+def projection(
+    d: 'Dict[str, _V]',
+    keys: 'Union[Dict[str, str], Iterable[str]]'
+) -> 'Dict[str, _V]':
     """
     Return a new :class:`dict` with keys of ``d`` restricted to values in ``keys``.
 
--- a/st3/sublime_lib/_util/collections.py
+++ b/st3/sublime_lib/_util/collections.py
@@ -1,10 +1,22 @@
 from collections.abc import Mapping, Sequence
 
+try:
+    from typing import Callable, Dict, Iterable, TypeVar, Union, overload
+
+    _V = TypeVar('_V')
+    _Result = TypeVar('_Result')
+    _Default = TypeVar('_Default')
+
+    _Value = Union[bool, int, float, str, list, dict, None]
+except ImportError:
+    pass
+    overload = lambda function: function
+
 
 __all__ = ['projection', 'get_selector', 'isiterable', 'ismapping', 'is_sequence_not_str']
 
 
-def projection(d, keys):
+def projection(d: 'Dict[str, _V]', keys: 'Union[Dict[str, str], Iterable[str]]') -> 'Dict[str, _V]':
     """
     Return a new :class:`dict` with keys of ``d`` restricted to values in ``keys``.
 
@@ -35,7 +47,31 @@ def projection(d, keys):
         }
 
 
-def get_selector(selector, default_value=None):
+@overload
+def get_selector(
+    selector: 'Callable[[Dict[str, _Value]], _Result]',
+    default_value: object = None
+) -> 'Callable[[Dict[str, _Value]], _Result]':
+    ...
+
+
+@overload  # noqa: F811
+def get_selector(
+    selector: str,
+    default_value: '_Default' = None
+) -> 'Callable[[Dict[str, _Value]], Union[_Value, _Default]]':
+    ...
+
+
+@overload  # noqa: F811
+def get_selector(
+    selector: 'Iterable[str]',
+    default_value: object = None
+) -> 'Callable[[Dict[str, _Value]], Dict[str, _Value]]':
+    ...
+
+
+def get_selector(selector, default_value=None):  # noqa: F811
     if callable(selector):
         return selector
     elif isinstance(selector, str):
@@ -48,17 +84,17 @@ def get_selector(selector, default_value=None):
         )
 
 
-def isiterable(obj):
+def isiterable(obj: object) -> bool:
     try:
-        iter(obj)
+        iter(obj)  # type: ignore
         return True
     except TypeError:
         return False
 
 
-def ismapping(obj):
+def ismapping(obj: object) -> bool:
     return isinstance(obj, Mapping)
 
 
-def is_sequence_not_str(obj):
+def is_sequence_not_str(obj: object) -> bool:
     return isinstance(obj, Sequence) and not isinstance(obj, str)
--- a/st3/sublime_lib/_util/collections.py
+++ b/st3/sublime_lib/_util/collections.py
@@ -1,6 +1,6 @@
 from collections.abc import Mapping, Sequence
 
-from .._compat.typing import Callable, Dict, Iterable, TypeVar, Union, overload
+from .._compat.typing import Callable, Dict, Iterable, TypeVar, Union
 
 
 _V = TypeVar('_V')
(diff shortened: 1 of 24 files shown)


st3/sublime_lib/settings_dict.py
```diff
@@ -108,18 +115,23 @@ def pop(self, key, default=_NO_DEFAULT):
...
-    def setdefault(self, key, default=None):
+    def setdefault(self, key: str, default: Value = None) -> Value:
         """Set `self[key]` to `default` if it wasn't already defined and return `self[key]`.
         """
         if key in self:
             return self[key]
         else:
-            self[key] = default
+            # self[key] = default
+            self.__setitem__(key, default)
```
FichteFoll: Why's this change necessary?

st3/sublime_lib/show_selection_panel.py
```diff
@@ -10,16 +16,16 @@
...
     *,
-    flags=0,
-    labels=None,
-    selected=NO_SELECTION,
-    on_select=None,
-    on_cancel=None,
-    on_highlight=None
-):
+    flags: Any = 0,
+    labels: Union[_Sequence[object], Callable[[_ItemType], object]] = None,
```
FichteFoll: More `Any` here?

st3/sublime_lib/show_selection_panel.py
```diff
@@ -85,17 +91,17 @@ def show_selection_panel(
     elif len(items) != len(labels):
         raise ValueError("The lengths of `items` and `labels` must match.")
 
-    def normalize_label(label):
-        if is_sequence_not_str(label):
+    def normalize_label(label: object) -> List[str]:
```
FichteFoll: Any

st3/sublime_lib/view_stream.py
```diff
@@ -56,27 +59,29 @@ def guard_auto_indent(self):
...
+        self, view: sublime.View, *, force_writes: bool = False, follow_cursor: bool = False
+    ):
         self.view = view
         self.force_writes = force_writes
         self.follow_cursor = follow_cursor
 
     @guard_validity
     @guard_selection
-    def read(self, size):
+    def read(self, size: int) -> str:
```
FichteFoll:
```suggestion
    def read(self, size: Optional[int]) -> str:
```

st3/sublime_lib/view_stream.py
```diff
@@ -88,7 +93,7 @@ def read(self, size):
 
     @guard_validity
     @guard_selection
-    def readline(self, size=-1):
+    def readline(self, size: int = -1) -> str:
```
FichteFoll: I just now noticed that we don't have the `size` parameter documented. Also seems weird to default to `-1` instead of `None`.

st3/sublime_lib/view_stream.py
```diff
@@ -98,7 +103,7 @@ def readline(self, size=-1):
 
         return self._read(begin, end, size)
 
-    def _read(self, begin, end, size):
+    def _read(self, begin: int, end: int, size: int) -> str:
```
FichteFoll:
```suggestion
    def _read(self, begin: int, end: int, size: Optional[int]) -> str:
```

st3/sublime_lib/view_utils.py
```diff
@@ -104,7 +108,7 @@ def close_view(view, *, force=False):
         raise ValueError('The view could not be closed.')
 
 
-def validate_view_options(options):
+def validate_view_options(options: Any) -> None:
```
FichteFoll:
```suggestion
def validate_view_options(options: Mapping[str, Any]) -> None:
```

st3/sublime_lib/view_utils.py
```diff
@@ -117,18 +121,19 @@ def validate_view_options(options):
...
+    name: Optional[str] = None,
+    settings: Optional[dict] = None,
+    read_only: Optional[bool] = None,
+    scratch: Optional[bool] = None,
+    overwrite: Optional[bool] = None,
+    syntax: Optional[str] = None,
+    scope: Optional[str] = None,
+    encoding: Optional[str] = None,
+    content: Optional[str] = None,
+    line_endings: Optional[str] = None
```
FichteFoll: I'm not sure if we do this in the other places, but I prefer always including trailing commas in multi-line collections for cleaner diffs.

st3/sublime_lib/_util/collections.py
```diff
@@ -35,30 +46,30 @@ def projection(d, keys):
         }
 
 
-def get_selector(selector, default_value=None):
+def get_selector(selector: object, default_value: object = None) -> Callable:  # noqa: F811
```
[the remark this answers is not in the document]
Thom1729: `object` is the top type, whereas `Any` disables type checking on that value. `object` is appropriate here because we can't do anything to the args until we've narrowed the type down manually in the function.

Most occurrences of `Any` in this PR should probably be removed, though it's probably not a blocker.

st3/sublime_lib/settings_dict.py
Thom1729: Reverted.

st3/sublime_lib/show_selection_panel.py
Thom1729: `object` is correct; we're not doing anything to that value except calling `str`, which is allowed on any value.

st3/sublime_lib/show_selection_panel.py
Thom1729: Same as above.

st3/sublime_lib/view_stream.py
Thom1729: Definitely should document. `-1` is used in the `io` classes. I would prefer `None`, but I think it's more important to maintain consistency.

st3/sublime_lib/view_stream.py
Thom1729: Should be `size: int = -1`.

st3/sublime_lib/view_stream.py
Thom1729: This can just be a mandatory `int` with no default because we're always passing a `int`.

st3/sublime_lib/view_utils.py
Thom1729: Fixed.

st3/sublime_lib/view_utils.py
Thom1729: Alas, Python doesn't allow a final trailing comma for keyword-only arguments. No idea why not, but it's a syntax error.

st3/sublime_lib/show_selection_panel.py
FichteFoll: Well, yeah, but isn't Any the same then? All things in Python are objects.

st3/sublime_lib/_util/collections.py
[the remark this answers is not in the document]
FichteFoll: Thanks, this comment prompted me to do some research and [the explanation in PEP 484](https://www.python.org/dev/peps/pep-0484/#the-any-type) made things quite clear. While both `Any` and `object` apply to all types, one is the higher and the other is the lower end of abstraction. 

Since we only perform operations onto what a basic object can do and don't have any further restrictions, we want `object` here.

st3/sublime_lib/view_stream.py
FichteFoll: I agree with that.

st3/sublime_lib/view_stream.py
FichteFoll: In that case we can remove the `None` check below.

st3/sublime_lib/view_utils.py
FichteFoll: Ah well, it's only been added in 3.6, sadly.

st3/sublime_lib/view_stream.py
```diff
@@ -94,7 +94,7 @@ def read(self, size: int) -> str:
     @guard_validity
     @guard_selection
     def readline(self, size: int = -1) -> str:
-        """Read until newline or EOF and return a single :class:`str`.
+        """Read and return one line from the stream. to a maximum of `size` characters.
```
FichteFoll: Extraneous `.` here.

Thom1729:
Addressed the above comments.

PR #130, opened 2019-08-25
Add .DS_Store to .gitignore.

Thom1729:


FichteFoll:
I'd prefer this kind of file to be added to some sort of default global ignore list (I do have it in mine, iirc), but sadly there is no default global ignore list in git.

PR #131, opened 2019-08-25
Implement __eq__ for SettingsDict

### LINKED ISSUE #129: Two SettingsDict should be equal if they refer to the same internal settings
A `sublime.Settings` object has a numeric `settings_id` that it passes to the internal API. Two `Settings` with the same `settings_id` are interchangeable for all practical purposes. By extension, the same is true for `SettingsDict` objects. Should `SettingsDict` reflect this?:

```
def __eq__(self, other):
    return isinstance(other, SettingsDict) and self.settings.settings_id == other.settings.settings_id
```

Under this implementation, `SettingsDict` and `NamedSettingsDict` objects could compare equal. I don't have a strong opinion on whether they should; perhaps not, since they're not interchangeable objects.

(Also: `__hash__`?)

Thom1729:
This implementation checks type, so a base SettingsDict will never equal a NamedSettingsDict.

Closes #129.


st3/sublime_lib/settings_dict.py
```diff
@@ -49,6 +49,15 @@ def __iter__(self) -> NoReturn:
         """Raise NotImplementedError."""
         raise NotImplementedError()
 
+    def __eq__(self, other) -> bool:
+        """Return ``True`` if `self` and `other` are of the same type
+        and refer to the same underlying settings data.
+        """
+        return (
+            type(self) == type(other) and
+            self.settings.settings_id == other.settings.settings_id
```
FichteFoll: Huh, do we not use line breaks before binary operators?

Thom1729: Fixed.

FichteFoll (approved):

PR #144, opened 2021-02-21
Run ActivityIndicator on main thread.

### LINKED ISSUE #140: Request for a threadless activity indicator
The current ActivityIndicator spawns a thread. There is some noticeable overhead associated with this thread. I'm noticing some stuttering when the indicator starts. Would it be possible to have a design that utilizes the assumption that the main thread is never blocked?

Thom1729:
Fix #140.

Might arguably be slightly more complicated than using a thread, but lower overhead, which matters in practice (e.g. https://github.com/sublimelsp/LSP-pyright/issues/36).


st3/sublime_lib/activity_indicator.py
```diff
@@ -102,30 +115,32 @@ def start(self) -> None:
...
         :raise ValueError: if the indicator is already running.
         """
-        with self._running:
-            if self._running.state:
+        with self._callback:
+            if self._callback.state is not None:
                 raise ValueError('Timer is already running')
             else:
-                self._running.state = True
+                callback = CancellableCallback(self._run)
```
FichteFoll, line 122: Why go the extra mile of wrapping the callable if the entire state is bound to an AcitvityIndicator instance anyway?

Thom1729: We have to cancel the callback somehow in order to implement `stop`. The `lsp_utils` implementation tracks this via a single `ActivityIndicator._state` variable, but this leads to a bug: if you `stop` and then `start` again before the already-scheduled timeout would have run, it will start schedule a new timeout *and* the original will run anyway, so now it's running at double speed. (I tested this and you can get it to run arbitrarily fast.)

Whatever action we take in `stop()` has to be somehow tied to the specific matching invocation of `start()`. A cancellable callback is a way of doing this. There are probably other ways of doing it, but a cancellable callback seemed to be the most straightforward.

Thom1729: (Offhand, another way we could do this is by having a `ActivityIndicator._invocation_id`. Something like:

```python
_invocation = 0

def start(self):
  sublime.set_timeout(partial(self._run, self._invocation), 100)

def _run(self, id):
  if id == self._invocation:
    self.tick()
    sublime.set_timeout(partial(self._run, id), 100)

def stop(self):
  self._invocation += 1
```

(Thread-safety omitted.) This would use less state, but it might be less intuitive than cancelling the callback.)

FichteFoll: Took quite a while to get to it, but I proposed an alternative in #150.

Thom1729:
I went with the invocation ID implementation to preserve existing behavior. It probably doesn't matter much, but it's something.

PR #148, opened 2021-02-23
Add delay to window_util tests.

Thom1729:
Due to https://github.com/sublimehq/sublime_text/issues/3960. If that's fixed, we can roll this back.


FichteFoll:
Merging as a part of #149.

PR #149, opened 2021-02-23
Revert scratchness after closing view or window

### LINKED ISSUE #147: Possible data loss bug in close_view and close_window, plus new clone_view
When using `close_view` or `close_window` with `force=True`, when a view is dirty and not scratch, the function will make the view scratch before closing it. But because scratchness is a buffer property, if there were other clones of that view, they will also be made scratch after the function is done. This could indirectly result in data loss.

The remedy is to keep track of what buffers `close_view` and `close_window` set to scratch and revert them afterward. I think that the original behavior is a bug, and the change in behavior is justified as a fix.

Anyway, when testing this, we need to create our own cloned views. Apparently the API doesn't provide anything like `view.clone()` — you have to:

1. Activate the view you want to clone.
2. Send the `clone_file` command to the window.
3. Return the active view.

This seems overcomplicated and (possibly) prone to race conditions. I wonder if there's a better way to do it. I also wonder whether a function that does it whatever way is best would be a good `sublime_lib` feature. See https://github.com/sublimehq/sublime_text/issues/3963.

Thom1729:
Fix #147.

Includes #148 so the window_util tests will run.

One case is not tested. If you have an unsaved view in one window and a clone of that view in the second window, and you `close_window()` the second window, the first view should not be scratch afterward. However, it doesn't look like it's possible to set that up using the Sublime API (https://github.com/sublimehq/sublime_text/issues/3965). I tested it by hand and It Works On My Machine™.

This PR does not expose `clone_view`.

### DIFF
--- a/st3/sublime_lib/view_utils.py
+++ b/st3/sublime_lib/view_utils.py
@@ -13,7 +13,6 @@ from ._compat.typing import Any, Optional, Mapping, Iterable, Generator
 
 __all__ = [
     'LineEnding', 'new_view', 'close_view',
-    '_clone_view', '_temporarily_scratch_unsaved_views',
 ]
 
 
--- a/st3/sublime_lib/view_utils.py
+++ b/st3/sublime_lib/view_utils.py
@@ -13,7 +13,7 @@ from ._compat.typing import Any, Optional, Mapping, Iterable, Generator
 
 __all__ = [
     'LineEnding', 'new_view', 'close_view',
-    'clone_view', 'temporarily_scratch_unsaved_views',
+    '_clone_view', '_temporarily_scratch_unsaved_views',
 ]
 
 
@@ -95,7 +95,7 @@ def new_view(window: sublime.Window, **kwargs: Any) -> sublime.View:
 
 
 @contextmanager
-def temporarily_scratch_unsaved_views(
+def _temporarily_scratch_unsaved_views(
     unsaved_views: Iterable[sublime.View]
 ) -> Generator[None, None, None]:
     buffer_ids = {view.buffer_id() for view in unsaved_views}
@@ -114,7 +114,7 @@ def temporarily_scratch_unsaved_views(
         view.set_scratch(False)
 
 
-def clone_view(view: sublime.View) -> sublime.View:
+def _clone_view(view: sublime.View) -> sublime.View:
     window = view.window()
     if window is None:  # pragma: no cover
         raise ValueError("View has no window.")
@@ -142,7 +142,7 @@ def close_view(view: sublime.View, *, force: bool = False) -> None:
         if not force:
             raise ValueError('The view has unsaved changes.')
 
-        with temporarily_scratch_unsaved_views([view]):
+        with _temporarily_scratch_unsaved_views([view]):
             closed = view.close()
     else:
         closed = view.close()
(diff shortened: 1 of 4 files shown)


tests/test_window_utils.py
```diff
@@ -8,6 +8,7 @@
 class TestNewWindow(DeferrableTestCase):
 
     def tearDown(self):
+        yield 100
```
FichteFoll, line 11: Could you add a comment why the wait is necessary here and reference the upstream issue? We won't be able to remove this for a while until we drop compatibility with affected builds.

st3/sublime_lib/view_utils.py
```diff
@@ -90,6 +94,40 @@ def new_view(window: sublime.Window, **kwargs: Any) -> sublime.View:
...
+        view.buffer_id(): view
+        for window in sublime.windows()
+        for view in window.views()
+        if view.buffer_id() in buffer_ids
+    }
+    for view in clones.values():
+        view.set_scratch(False)
+
+
+def clone_view(view: sublime.View) -> sublime.View:
```
FichteFoll, line 117: Do we intend to expose this? If yes, it should receive a docstring.

st3/sublime_lib/view_utils.py
```diff
@@ -90,6 +94,40 @@ def new_view(window: sublime.Window, **kwargs: Any) -> sublime.View:
     return view
 
 
+@contextmanager
+def temporarily_scratch_unsaved_views(
```
FichteFoll, line 98: Do we intend to expose this? If no, it should probably begin with `_` and not be included in `__all__`.

FichteFoll (changes requested):

st3/sublime_lib/view_utils.py
Thom1729: We do need to export it here because it's used in `window_utils.close_window` (just as `clone_view` is used in the unit tests). I don't have an opinion about the underscore — given that we do have to export it, would you rather have the underscore? If so, I'll add it to both `temporarily_scratch_unsaved_views` and `clone_view`.

tests/test_window_utils.py
Thom1729: Done.

st3/sublime_lib/view_utils.py
FichteFoll: An underscore means its for internal use and specifically not included in star-imports, so yes, please do. You can't not expose something in Python anyway (comparable to a "private" visibility modifier) and underscore is closest to a "protected", which is what we want here.

Thom1729: Fixed.

st3/sublime_lib/view_utils.py
```diff
@@ -13,7 +13,7 @@
 
 __all__ = [
     'LineEnding', 'new_view', 'close_view',
-    'clone_view', 'temporarily_scratch_unsaved_views',
```
FichteFoll: Exporting an underscored name in `__all__` kind of defeats the purpose of underscoring it in the first place, doesn't it? Those should be mutually exclusive. 

The functions are for internal use only and we'll never use a star import in our own code, I presume, making this redundant.

Thom1729: I know it doesn't do anything, but for some reason I thought that mypy used it as a hint to check imports. Apparently it doesn't, so I've removed them from `__all__`.

PR #150, opened 2021-03-01
Alternative for ActivityIndicator on main thread

FichteFoll:
Implement the cancellation logic inside the scheduling class instead of
a different single-purposed class with a single additional boolean that
allows cancelling (or reverting) a cancellation.


FichteFoll:
I tried running the tests locally, but there appears to be a problem with UnitTesting in my setup. Looks like tests passed on CI anyway.

st3/sublime_lib/activity_indicator.py
```diff
@@ -95,8 +86,7 @@ def __init__(
             self._target = target
 
         self._ticks = 0
-
-        self._callback = LockedState(None)
+        self._locked_state = SchedulerState()
```
Thom1729, line 89: Should be `self._locked_state = LockedState(SchedulerState())`.

Thom1729 (commented):
Unfortunately, this is vulnerable to a race condition.

```py
from sublime_lib import ActivityIndicator; a = ActivityIndicator(window); a.start(); a.stop(); a.start()
# ValueError: Timer is already running
```

Because there's no (native) way to cancel a set_timeout callback, we need to keep track of some value particular to that invocation of set_timeout — which means keeping track of multiple values at the same time, because we can stop/start repeatedly before any of the callbacks have run.

If an entire `CancellableCallback` class seems overcomplicated, I can try to implement the `invocation_id` concept to see how that looks.

FichteFoll:
Nice find. I did think about this case, but missed it in one condition.

There can only be 3 states:

- nothing is running
- a timeout is scheduled and it should run
- a timeout is scheduled and it should not run

We can totally switch between these three cases if we consider them properly.

Thom1729:
This will result in different behavior from the threaded implementation. Suppose that we start, stop, wait 50 milliseconds, and start again. The callback should run at t=150, 250, …, but it will actually run at t=100, 200, … because it will “re-use” the callback from the first start rather than creating a new one timed from the second (most recent) start.

There are actually infinitely many possible states:

- *n* ≥ 1 timeouts are scheduled and the most recent should run.
- *n* ≥ 0 timeouts are scheduled and the most recent should not run.

FichteFoll:
> The callback should run at t=150, 250, …, but it will actually run at t=100, 200

I don't see a problem with that. In the context of showing an acitvity indicator in the status bar, the user will not care whether the update happens 50ms earlier or later and neither does the developer. What they care about is the indicator being shown at all and updated at a consistent interval.

PR #151, opened 2021-03-01
Use invocation ID to track most recent callback instead of cancellable callbacks.

Thom1729:
This implements the solution I suggested in the comments on #144. Each time `stop()` is called, if the ActivityIndicator is running, it will increment a counter. Each callback knows (via `partial()`) what the value of the counter was when it was scheduled, so when the callback runs it can tell whether `stop()` was called in the meantime.

So we have mutable state (the counter) and per-start state (the value of the counter when it was started), but no mutable per-start state.

This PR also removes LockedState. The original motivation for LockedState was so that we could use `Condition.wait_for()` to detect when the ActivityMonitor was stopped. Since we're not using threading anymore, a simple Lock works just fine. We can keep LockedState around if we want, but personally I don't think it's worth it.


FichteFoll:
I have adjusted my PR to also remove LockedState, so that they we can compare them better.

I definitly like this version much better than what we currently have in #144. I don't have a strong opinion on whether I prefer it over #150, however, so I'll leave that decision to you. Consider the reviews of all 3 as "approved".

PR #152, opened 2021-03-29
Fix a bug when changing settings in a subscribe callback.

Thom1729:
Before this change, the saved value of a setting wasn't updated until after the callback was completed, so if you changed that settings object inside the callback, it would incorrectly detect that the first setting had changed again.

### DIFF
--- a/st3/sublime_lib/settings_dict.py
+++ b/st3/sublime_lib/settings_dict.py
@@ -181,15 +181,16 @@ class SettingsDict():
         """
         selector_fn = get_selector(selector)
 
-        previous_value = selector_fn(self)
+        saved_value = selector_fn(self)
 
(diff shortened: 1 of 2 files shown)


FichteFoll:
This fixes an error case, but I still see potential for confusion when a "complex" object is constructed by the selector and modified in place, since the callback is called by reference, but that is a slightly different issue. To tackle that, we'd probably need a mutex and a second call of the selector to ensure we get two copies of the same state without relying on the data to be copyable, which is definitely not ideal.

Besides, I don't think it is currently documented that the selector's result must be equatable.

Thom1729:
That's a good point. I think it could be handled separately from this bug, though.

FichteFoll (approved):

PR #154, opened 2021-04-01
Improve subscribe documentation.

Thom1729:
For https://github.com/SublimeText/sublime_lib/pull/152#issuecomment-810678572.

My feeling is that mutating the selector result is kind of a weird thing to do and that it's probably not worth doing something complicated with mutexes.


FichteFoll:
> My feeling is that mutating the selector result is kind of a weird thing to do

Definitely. Just documenting it seems fine to me.

FichteFoll (approved):

Issue #134, opened 2020-03-03
ResourcePath.copytree fails for directory inside compressed package


rchl:
I have package `LSP-eslint.sublime-package` with file structure like:
```
/vscode-eslint/
/vscode-eslint/server.js
/vscode-eslint/more_files
/plugin.py
```
I'm trying to copy whole `vscode-eslint` directory to cache directory with:

```py
source_resource_path = 'Packages/LSP-eslint/vscode-eslint/'
target_path = os.path.join(sublime.cache_path(), 'LSP-eslint', 'server')
sublime_lib.ResourcePath(source_resource_path).copytree(target_path, exist_ok=True)
```

That fails with:
```
Traceback (most recent call last):
  File "/Users/me/Library/Application Support/Sublime Text 3/Packages/sublime_lib/st3/sublime_lib/resource_path.py", line 521, in read_bytes
    return sublime.load_binary_resource(str(self))
  File "/Applications/Sublime Text.app/Contents/MacOS/Lib/python33/sublime.py", line 244, in load_binary_resource
    raise IOError('resource "%s" not found' % name)
OSError: resource "Packages/LSP-eslint/vscode-eslint" not found

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/Applications/Sublime Text.app/Contents/MacOS/Lib/python33/sublime_plugin.py", line 478, in on_api_ready
    plc()
  File "/Users/me/Library/Application Support/Sublime Text 3/Installed Packages/LSP-eslint.sublime-package/plugin.py", line 52, in plugin_loaded
    sublime_lib.ResourcePath(local_server_path).copytree(server_cache_directory_path, exist_ok=True)
  File "/Users/me/Library/Application Support/Sublime Text 3/Packages/sublime_lib/st3/sublime_lib/resource_path.py", line 608, in copytree
    resource.copy(file_path)
  File "/Users/me/Library/Application Support/Sublime Text 3/Packages/sublime_lib/st3/sublime_lib/resource_path.py", line 582, in copy
    data = self.read_bytes()
  File "/Users/me/Library/Application Support/Sublime Text 3/Packages/sublime_lib/st3/sublime_lib/resource_path.py", line 523, in read_bytes
    raise FileNotFoundError(str(self)) from err
FileNotFoundError: Packages/LSP-eslint/vscode-eslint
```

It works fine when package is unpacked.
[...]

PR #157, opened 2021-04-04
Fix glob bug.

Thom1729:
For #134.

After much research and testing, I determined that the fundamental bug was that every pattern component except for `**` should match at least one character — so `/A*B/` should match `AB`, but `A/*` shouldn't match `A/`. This behavior should match other glob engines and resolve the linked bug.

This behavior varies slightly from `sublime.find_resources('*')`, which will find all resources (like `sublime.find_resources('')`.

It's probably bad news for other reasons if `sublime.find_resources()` returns directories, but at least now it shouldn't be bad news for us in particular.

### DIFF
--- a/st3/sublime_lib/_util/glob.py
+++ b/st3/sublime_lib/_util/glob.py
@@ -28,7 +28,7 @@ def get_glob_matcher(pattern: str) -> Callable[[str], bool]:
             # Component must not be empty.
             expr_string += r'(?:[^/])+' + '/'
         elif component == '**':
-            expr_string += r'(?:(?:.*)(?:\Z|/))?'
+            expr_string += r'(?:.*(?:\Z|/))?'
         elif '**' in component:
             raise ValueError("Invalid pattern: '**' can only be an entire path component")
         else:
--- a/st3/sublime_lib/_util/glob.py
+++ b/st3/sublime_lib/_util/glob.py
@@ -7,8 +7,7 @@ __all__ = ['get_glob_matcher']
 
 
 GLOB_RE = re.compile(r"""(?x)(
-    (?:^|/) \*\* (?:$|/)
-    | (?<!\*)\*(?!\*)
+    \*
     | \?
     | \[ .*? \]
 )""")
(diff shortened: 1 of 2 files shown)


Thom1729:
N.B. In order to directly test #134, we'd have to create a zipfile with directory entries, which I'm not sure how to do within Python.

FichteFoll:
> N.B. In order to directly test #134, we'd have to create a zipfile with directory entries, which I'm not sure how to do within Python.

Yeah, I tried writing such a file with the zipfilemodule, but you can't create "files" with trailing slashes.
I suppose we need to bundle such a zip as an artifact.

st3/sublime_lib/_util/glob.py
```diff
@@ -7,36 +7,45 @@
...
-        elif part == '?':
-            s += r'(?:[^/])'
-        elif part[0] == '[':
-            s += part
-        elif '**' in part:
+        elif component == '*':
+            # Component must not be empty.
+            expr_string += r'(?:[^/])+' + '/'
+        elif component == '**':
+            expr_string += r'(?:(?:.*)(?:\Z|/))?'
```
FichteFoll, line 31:
```suggestion
            expr_string += r'(?:.*(?:\Z|/))?'
```

tests/test_glob.py
```diff
@@ -115,6 +115,7 @@ def test_recursive(self):
         self._test_matches(
             'Foo/**',
             [
+                'Packages/Foo/',
```
FichteFoll, line 118: Should we match this at all?

Thom1729: Yes, because a UNIX glob would and I don't think we need to deviate. It's a corner case that probably won't come up very often, but if it does then I think handling it like a UNIX glob is the least surprising thing to do.

Thom1729:
> I suppose we need to bundle such a zip as an artifact.

Ugh. You're probably right, but still ugh. I've [opened a question](https://stackoverflow.com/questions/66970927/create-zipfile-with-directory-entries-in-python) on Stack Overflow looking for an alternative.

Thom1729:
Do you think this can be merged as-is, punting the direct zipfile test to a future PR?

FichteFoll (commented):
Yeah, we can merge this.

PR #161, opened 2021-04-10
Factor out weak_method.

Thom1729:
Factor out weak_method so that it can be used elsewhere and add a basic test.

It might be useful to offer something like this as a utility function for users to pass callbacks to e.g. `sublime.set_timeout` without unnecessarily keeping things alive.

### DIFF
--- a/st3/sublime_lib/_util/weak_method.py
+++ b/st3/sublime_lib/_util/weak_method.py
@@ -1,0 +1,21 @@
+import weakref
+
+from .._compat.typing import Callable, Any
+from types import MethodType
+
+
+__all__ = ['weak_method']
+
+
+def weak_method(method: Callable) -> Callable:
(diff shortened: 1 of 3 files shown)


FichteFoll (approved):
FWIW, I'm using something similar in PackageDev, so yes, this definitely has use. I know we aren't exposing it at the moment and probably shouldn't, because Python 3.4 added `weakref.WeakMethod`, which has slightly different semantics.

Thom1729:
I considered WeakMethod, but this does something that's slightly (but usefully) different — it returns an ordinary function that you can pass right into e.g. `set_timeout`. You can't pass a WeakMethod; you'd have to wrap it anyway. And it's slightly different from a `weakref.Proxy` because it won't raise a `ReferenceError`.

It's not that much of a difference, and YMMV on whether it makes sense to expose.

Issue #136, opened 2020-04-10
input_handler class decorator to automatically add and create simple input handlers


FichteFoll:
I was just thinking that most (text) input handlers are fairly simple and provide mostly static values for their (abstract) methods, which makes creation of a separate class rather decoupled from the command's not as transparent as it needs to.
Furthermore, implementing the command's `input` seems like busy work most of the time and things become especially complicated once you start using multiple handlers and want them to chain using `next_input`.

I believe a decorator in `ArgumentParser.add_argument` style could make this API more streamlined while still allowing for the complex case.

API example:
```py
@input_handler("param", kind="text", placeholder="Replace me", validate=lambda _: True)
@input_handler("param2", kind="list", items=list(range(10)))
@input_handler(Param3InputHandler)  # name *can* be determined from handler's `name()`
class TestCommand(sublime_plugin.WindowCommand):
    def run(self, explicit_param, param, param2, param3, opt_param=None):
        pass
```

## Questions

1. Technically, input handlers don't have access to the command instance, the window/view or any of the previous parameters. Especially for the complex case, I'm unsure how to provide this information since you'd use custom handlers mostly in those situations when that information is needed. How to handle this? Just pass `window/view` and a dict of all args to the handler's `__init__`?

1. The same issue also surfaces when the list items you want to provide are dynamic.

1. For a similar reason, I don't think `confirm`, `cancel`, or `description` make much sense to be supplied in the decorator short-hand form.

1. Use the `kind` param to distinguish, decide based on an `items` parameter, or use two decorators use `text_input_handler` and `list_input_handler`?

FichteFoll:
If you have some complex input handler examples, I believe trying to wrap those into such a generator as an example would help with designing the API for this.

Thom1729:
I've played around with this recently. The stumbling block for me is that I can't really think of any especially complex examples out in the wild. (Maybe this is a chicken-and-egg thing, and more people would write input handlers if they were easier to work with.)

So I wrote up a quick example implementation of something like PackageResourceViewer's Open Resource command:

```py
class InputTestCommand(sublime_plugin.WindowCommand):
    def run(self, path):
        content = sublime.load_resource(path)
        new_view(
            self.window,
            name=path,
            content=content,
            syntax=sublime.find_syntax_for_file(path),
            read_only=True,
            scratch=True,
        )

    @fancy_input
    def input(self, args):
        path = ResourcePath('Packages')

        while not path.exists():
[...]

PR #163, opened 2021-04-10
New TemporaryPackage utility for tests.

Thom1729:
To abstract out the process of creating/cleaning up a temporary package. We'll use this for testing #136, at the very least.

### DIFF
--- a/st3/sublime_lib/_util/temporary_package.py
+++ b/st3/sublime_lib/_util/temporary_package.py
@@ -1,0 +1,30 @@
+import shutil
+from ..resource_path import ResourcePath
+
+
+__all__ = ['TemporaryPackage']
+
+
+class TemporaryPackage:
+    def __init__(self, package_name: str, resource_path: ResourcePath) -> None:
+        self.package_name = package_name
+        self.resource_path = resource_path
+
+    @property
+    def package_path(self) -> ResourcePath:
+        return ResourcePath("Packages") / self.package_name
+
+    def create(self) -> None:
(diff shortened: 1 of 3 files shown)


st3/sublime_lib/_util/temporary_package.py
```diff
@@ -0,0 +1,30 @@
+import shutil
```
FichteFoll, line 1: A module or class docstring for this would be nice, since its use cases isn't immediately obvious compared to the other parts of this library, imo.

tests/test_resource_path.py
```diff
@@ -1,28 +1,26 @@
 import sublime
-import shutil
 import tempfile
 
 from sublime_lib import ResourcePath
 from sublime_lib._compat.pathlib import Path
+from sublime_lib._util.temporary_package import TemporaryPackage
```
FichteFoll, line 6: If it's only going to be used for tests, we should move it to the tests package.

Thom1729: I'm probably missing something obvious, but I couldn't get UnitTesting to import it because of the weird way that `unittest` loads tests. Is there a simple way around this? I'd definitely move it out to the tests directory if possible.

FichteFoll: We don't have a `__init__.py` in the tests folder. We should be able to use relative imports with that.

Thom1729: Done.

FichteFoll (approved):

Issue #141, opened 2020-12-30
A "maintainer" or "owner" class for view.add_regions


rwols:
Decorated regions in a view can be regarded as a resource. Resources must be cleaned up when they are no longer needed.

It is too easy to forget this in a large plugin (or even a small plugin). For instance, one tends to forget to clean up regions when a plugin gets disabled.

It'd be nice to have a class that maintains this resource. Something like this (just an idea):

```python
class AddRegionsMaintainer:

    def __init__(self, view: sublime.View, key: str, *args: Any, **kwargs: Any) -> None:
        self._view = view
        self._key = key
        self(*args, **kwargs)

    def __call__(self, *args: Any, **kwargs: Any) -> None:
        self._view.add_regions(self._key, *args, **kwargs)

    def __del__(self) -> None:
        self._view.erase_regions(self._key)
```

Thom1729:
@rwols The linked PR uses the following interface:

```python
class RegionManager:
    def __init__(
        self,
        view: sublime.View,
        key: Optional[str] = None,
        *,
        scope: Optional[str] = None,
        icon: Optional[str] = None,
        flags: Optional[RegionOption] = None
    ): ...

    def __del__(self) -> None: ...

    def set(
        self,
        regions: Collection[sublime.Region],
        *,
        scope: Optional[str] = None,
        icon: Optional[str] = None,
        flags: Optional[RegionOption] = None
    ) -> None: ...

    def get(self) -> List[sublime.Region]: ...

    def erase(self) -> None: ...

```

Is it actually useful to override the scope/icon/flags in `set()`?

rwols:
I would say yes. If for instance the icon is determined by a package setting, you can just load that setting when updating the regions.

PR #165, opened 2021-04-11
Add RegionManager.

Thom1729:
For #141.

I think that most of the time, you'd be using the same scope, icon, etc for a set of regions, so it makes sense to put that in the constructor. You can still override that in `set()`.

Notes:

- New ST4 stuff not yet included.
- `View.get_regions()` doesn't give any details, so the only way to test the args is by mocking View. I have not yet done this.
- Rather than accepting a `flags` argument, maybe we should break this out by flag. That especially makes more sense when overriding defaults in `set()`.


st3/sublime_lib/region_manager.py
```diff
@@ -0,0 +1,61 @@
...
+        self.scope = scope
+        self.icon = icon
+        self.flags = flags
+
+    def __del__(self) -> None:
+        self.erase()
+
+    def set(
+        self,
+        regions: List[sublime.Region],
```
FichteFoll, line 43: The implementation in `sublime` requires annotations to be a list, but regions seemingly only need to have a length and be iterable, so a tuple also works (sets and dicts don't because `Region`s aren't hashable). I think the closest abstract match is a `Collection`.

st3/sublime_lib/region_manager.py
```diff
@@ -0,0 +1,61 @@
...
+from .flags import RegionOption
+
+
+__all__ = ['RegionManager']
+
+
+T = TypeVar('T')
+
+
+def coalesce(*values: Optional[T]) -> T:
```
FichteFoll, line 13: `_coalesce`?

st3/sublime_lib/region_manager.py
```diff
@@ -0,0 +1,61 @@
...
+        self.flags = flags
+
+    def __del__(self) -> None:
+        self.erase()
+
+    def set(
+        self,
+        regions: List[sublime.Region],
+        *,
+        scope: Optional[str] = None,
```
FichteFoll, line 45: Should the scope/icon/flags overrides of this function affect the manager's defaults? If not, what use case do we imagine for manually overriding these for a single `set`?

st3/sublime_lib/region_manager.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+T = TypeVar('T')
+
+
+def coalesce(*values: Optional[T]) -> T:
+    return next(value for value in values if value is not None)
+
+
+class RegionManager:
```
FichteFoll, line 17: No documentation yet.

st3/sublime_lib/region_manager.py
Thom1729: Fixed.

st3/sublime_lib/region_manager.py
Thom1729: The `_coalesce` arguments were backward; the arguments to `set` are supposed to supersede the defaults. I've fixed it.

st3/sublime_lib/region_manager.py
Thom1729: Do you think we should change the `sublime` type stub?

FichteFoll: Yeah.

st3/sublime_lib/region_manager.py
FichteFoll: Ah sure, I didn't even catch this. I mean the existance of the overrides in this context in general.

st3/sublime_lib/region_manager.py
Thom1729: Done.

st3/sublime_lib/region_manager.py
Thom1729: I don't have a particular use case in mind. Even without those arguments to `set`, you could always just change the attributes manually. I'll ping rwols in the issue to see if he has an opinion.

tests/test_region_manager.py
```diff
@@ -3,6 +3,24 @@
...
+        super().__init__()
+
+        def side_effect(*args, **kwargs):
+            bound_args = (signature(sublime.View.add_regions)
+                          .bind(self, *args, **kwargs).arguments)
+            callback(**bound_args)
+
+        self.add_regions = MagicMock(
+            spec=sublime.View.add_regions,
+            side_effect=side_effect if callback else None
```
FichteFoll, line 22: The `callback` arg doesn't seem to be used?

Thom1729: Whoops; that's leftover code. Removed.

st3/sublime_lib/region_manager.py
```diff
@@ -15,6 +15,15 @@ def _coalesce(*values: Optional[T]) -> T:
...
 
 class RegionManager:
+    """A manager for regions in a given :class:`sublime.View` with the same `key`.
+
+    If `key` is not given,
+    a unique identifier will be used.
+
+    If the `scope`, `icon`, and `flags` args are given,
+    then they will be used as default values for :meth:`set`.
+    """
```
FichteFoll, line 25: We need some `versionadded` directive here, right? And include it in `index.rst` ofc.

FichteFoll (approved):
Other than wrapping up the docs, LGTM.

st3/sublime_lib/region_manager.py
Thom1729: Done.

FichteFoll (approved):

Issue #164, opened 2021-04-11
Settings listener classes


Thom1729:
As seen on Discord. Motivating examples:

```python
from sublime_lib import GlobalSettingsListener, ViewSettingsListener, on_setting_changed

class TabSizeListener(ViewSettingsListener):
    @classmethod
    def is_applicable(cls, settings: sublime.Settings) -> bool:
        return settings.get('translate_tabs_to_spaces', False)

    @classmethod
    def applies_to_primary_view_only(cls) -> bool:
        return True

    @on_setting_changed('tab_size')
    def tab_size_changed(self, new_value: object, old_value: object) -> None:
        self.view.run_command('resize_existing_tabs', {
            'new_size': new_value,
            'old_size': old_value,
        })

class JsCustomConfigurationsListener(GlobalSettingsListener)
    SETTINGS_NAME = 'JS Custom.sublime-settings'

    @on_setting_changed(get_configurations)
    def configurations_changed(self, new_configurations, old_configurations):
        if not self.settings.get('auto_build', False):
            return

        changed = [
            name for name in set(old_configurations) | set(new_configurations)
            if old_configurations.get(name, None) != new_configurations.get(name, None)
        ]

        if changed:
            print('JS Custom: Configuration changed. Rebuilding some syntaxes.')
            sublime.active_window().run_command('build_js_custom_syntaxes', {'versions': changed})
```

This would provide the same convenience features as `SettingsDict.subscribe`: selecting the desired value from the settings, detecting whether that value has changed, and keeping track of the previous value. What is also provides beyond that is automatic management of the subscription lifetime. Manually calling `clear_on_change` is annoying, easy to forget, prone to mistakes, and when done right kind of boilerplate-y. A `SettingsListener` class can handle this automatically by simply inheriting from the existing listener classes (that Sublime already manages), calling `clear_on_change` in `__del__`, and using a weakref in `add_on_change`.

FichteFoll:
Do we want to ensure the decorator is only used on classes that subclass our SettingsListener classes or would that hinder mixin usage?

Can we even? I think at the time the decorator is called, the function hasn't been bound to the class yet, so that might be a moot point.

PR #166, opened 2021-04-12
Add SettingsListener.

Thom1729:
For #164.

Notes:

- Needs documentation, many more tests, more error handling.
- A couple of the types could be tightened up.
- `OnChangedOptions` is a named tuple so that we can easily support more arguments to `on_setting_changed`.
- One obvious idea is an `async` argument to `on_setting_changed`. Alternatively/additionally, we could have a separate `on_setting_changed_async` decorator.
- Instead of `SETTINGS_NAME = 'Foo'`, we could just have `settings = sublime.load_settings('Foo.sublime-settings')`. I think the former might be slightly less prone to user error.
- As you mentioned in the issue, it would be nice to have error handling if someone uses the `on_setting_changed` decorator outside a `BaseSettingsListener`. Haven't looked into implementation yet.
- If someone unintentionally exports `ViewSettingsListener` or `GlobalSettingsListener` from their plugin, Sublime will dutifully use them. This probably doesn't really matter, but we could special-case this if we want.
- Is there any reason to let users specify a custom key, rather than `str(id(self))`? I suspect not.
- Do we want a `WindowSettingsListener`? I also suspect not.
- Currently, `selector` only supports strings and functions, but we could just use `_util.collections.get_selector`.

### DIFF
--- a/st3/sublime_lib/settings_listener.py
+++ b/st3/sublime_lib/settings_listener.py
@@ -1,0 +1,85 @@
+import sublime
+import sublime_plugin
+
+from collections import namedtuple
+from ._util.weak_method import weak_method
+
+
+from ._compat.typing import Callable, Any, Iterator, Tuple, TypeVar, Union
+
+
+__all__ = ['GlobalSettingsListener', 'ViewSettingsListener', 'on_setting_changed']
+
+
+OnChangedOptions = namedtuple('OnChangedOptions', ('selector'))
+OPTIONS_ATTRIBUTE = '_sublime_lib_settings_listener_options'
+
+
+class BaseSettingsListener:
+    def _handlers(self) -> Iterator[Tuple[str, Callable, OnChangedOptions]]:
+        for name in dir(self):
+            value = getattr(self, name)
+            options = getattr(value, OPTIONS_ATTRIBUTE, None)
+
+            if not name.startswith('_') and options is not None:
+                yield name, value, options
+
+    def __init__(self, settings: sublime.Settings, *args: Any, **kwargs: Any) -> None:
+        # Don't complain that object.__init__ doesn't take any args
+        super().__init__(*args, **kwargs)  # type: ignore
+        self.settings = settings
+
+        self.settings.add_on_change(str(id(self)), weak_method(self._on_settings_changed))
+
+        self._last_known_values = {
+            name: options.selector(self.settings)
+            for name, _, options in self._handlers()
+        }
+
+    def __del__(self) -> None:
+        self.settings.clear_on_change(str(id(self)))
+
+    def _on_settings_changed(self) -> None:
+        for name, handler, options in self._handlers():
+            previous_value = self._last_known_values[name]
+            current_value = options.selector(self.settings)
+
+            if current_value != previous_value:
+                self._last_known_values[name] = current_value
+                handler(current_value, previous_value)
+
+
+class GlobalSettingsListener(BaseSettingsListener, sublime_plugin.EventListener):
+    SETTINGS_NAME = ''
+
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        super().__init__(sublime.load_settings(self.SETTINGS_NAME), *args, **kwargs)
+
+
+class ViewSettingsListener(BaseSettingsListener, sublime_plugin.ViewEventListener):
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        view = args[0]
+        assert isinstance(view, sublime.View)
+        super().__init__(view.settings(), *args, **kwargs)
+
+
+Selected = TypeVar('Selected')
+OnChangeListener = Callable[[BaseSettingsListener, Selected, Selected], None]
+
+
+def on_setting_changed(
+    selector: Union[str, Callable[[sublime.Settings], Selected]]
+) -> Callable[[OnChangeListener], OnChangeListener]:
+    if callable(selector):
+        selector_function = selector
+    elif isinstance(selector, str):
+        selector_str = selector
+        selector_function = lambda settings: settings.get(selector_str)
+    else:
+        raise TypeError('Selector must be a string or function.')
+
+    def decorator(function: OnChangeListener) -> OnChangeListener:
+        setattr(function, OPTIONS_ATTRIBUTE, OnChangedOptions(selector_function))
+        return function
+
+    return decorator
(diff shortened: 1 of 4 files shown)


Thom1729:
I've run into a bit of trouble with `GlobalSettingsListener`. It turns out that `EventListener`s are never garbage-collected. (I'm pretty sure this is a bug.)

I have solved the problem with a moderately ugly hack. `GlobalSettingsListener` implements an `on_new` handler that does nothing. Then, in `_on_settings_change`, it checks whether that callback is registered. If not, it calls its own `__del__`.

After spending several hours debugging this, I'm confident that this was the problem and that the solution works — though I wouldn't call it pretty.

(There's no such problem with `ViewEventListener`s.)

Thom1729:
@FichteFoll Any thoughts on the above hack?

st3/sublime_lib/settings_listener.py
```diff
@@ -57,6 +57,15 @@ class GlobalSettingsListener(BaseSettingsListener, sublime_plugin.EventListener)
     def __init__(self, *args: Any, **kwargs: Any) -> None:
         super().__init__(sublime.load_settings(self.SETTINGS_NAME), *args, **kwargs)
 
+    def _on_settings_changed(self) -> None:
+        if self in sublime_plugin.all_callbacks['on_new']:
+            super()._on_settings_changed()
+        else:
+            self.__del__()
```
FichteFoll, line 86: Pretty sure you're never supposed to call `__del__` yourself. If it's the functionality we want, move the code to a separate internal function.

st3/sublime_lib/settings_listener.py
```diff
@@ -52,12 +52,33 @@ def _on_settings_changed(self) -> None:
...
+
+            @on_setting_changed('configurations')
+            def configurations_changed(self, new_configuration, old_configuration):
+                if self.settings.get('auto_build', False):
+                    sublime.active_window().run_command('build_js_custom_syntaxes')
+    """
     SETTINGS_NAME = ''
 
     def __init__(self, *args: Any, **kwargs: Any) -> None:
         super().__init__(sublime.load_settings(self.SETTINGS_NAME), *args, **kwargs)
```
FichteFoll, line 78: I'd like to have some verification here that `self.SETTINGS_NAME` has been overridden in the subclass because that's easy to miss.

st3/sublime_lib/settings_listener.py
```diff
@@ -81,6 +127,21 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
...
+    A decorator function to mark a method
+    of a :class:`GlobalSettingsListener` or :class:`ViewSettingsListener`
+    as a setting change handler.
+
+    The handler is not called every time the `settings` object changes,
+    but only when the value derived using the `selector` argument changes.
+    If `selector` is callable, then the derived value is ``selector(self)``.
+    If `selector` is a :class:`str`,
+    then the derived value is ``self.get(selector, None)``.
+    Otherwise, the derived value is ``projection(self, selector)``.
```
FichteFoll, line 142: With the "selector" concept being reused now, it might be sensible to add a section to our docs with more details and examples.

FichteFoll (changes requested):

FichteFoll:
Regarding the `on_new` hack: I think it's reasonable, although I surely wish it wasn't necessary. Yet, I cannot think of a better idea, so LGTM in general.

st3/sublime_lib/settings_listener.py
Thom1729: Fixed.

st3/sublime_lib/settings_listener.py
Thom1729: I agree, though I'm not sure about the mechanism. Should we raise an error? Log a warning? Either way, it might be tricky to test.

We'll also want a special case for when someone unthinkingly exports `GlobalSettingsListener ` itself and Sublime instantiates it.

FichteFoll: I'd expect an exception, probably a `RuntimeError`.

And true, we need something to prevent the imported class from interfacing with ST.

Thom1729: Added.

FichteFoll (approved):
Nice, I think that was everything.

Edit: Oh, I forgot about the documentation thing. I'd still like to see that.

Thom1729:
I wrote up some docs for the projection function. I'm not completely satisfied with it, but it might be good enough.

st3/sublime_lib/settings_listener.py
```diff
@@ -145,13 +145,29 @@ def on_setting_changed(
 
     The handler is not called every time the `settings` object changes,
     but only when the value derived using the `selector` argument changes.
-    If `selector` is callable, then the derived value is ``selector(self)``.
+    If `selector` is callable, then the derived value is ``selector(settings)``.
     If `selector` is a :class:`str`,
```
FichteFoll, line 149: I believe this to be the most common use case, so it should be explained first. Either way, a callable is probably last.

st3/sublime_lib/settings_listener.py
```diff
@@ -145,13 +145,29 @@ def on_setting_changed(
...
-    then the derived value is ``self.get(selector, None)``.
-    Otherwise, the derived value is ``projection(self, selector)``.
+    then the derived value is ``settings.get(selector, None)``.
+    Otherwise, the derived value is ``projection(settings, selector)`` (as defined below).
 
     The handler should accept two arguments:
     the new derived value and the previous derived value.
+
+    ``projection(settings, keys)``
```
FichteFoll, line 156: Since we use selectors twice in `sublime_lib`, I really believe this explanation should go into its own section that we then reference. Kind of how the official docs have a [types section](https://www.sublimetext.com/docs/api_reference.html#types) that they link to for such special types.

Because `projection` isn't a function directly used by the user of this API and they can't change the first parameter, I believe it makes more sense to explain the effects of the second parameter only. I just don't think think the technical explanation is useful.

st3/sublime_lib/settings_listener.py
Thom1729: Good point; I've moved the string case first.

st3/sublime_lib/settings_listener.py
Thom1729: Sorry, my brain's been stuck in a loop on the `projection` docs. I agree there's room for improvement here, but is it good enough to release? If not, then it might be simplest to limit `@on_setting_changed` to a string or callable for now.

FichteFoll: Github doesn't show my comment to me anymore, so I'm adding another one just in case. 

It's probably good enough, but I added a separate issue to track this so we don't forget.

Also, kinda unrelated, but it would have been good if you removed the `[WIP]` tag in the PR title at some point. Since I've been struggling to find time for open source lately, I need to select carefully which issues and PRs I take a look into first.

PR #167, opened 2021-04-16
Various fixes for sublime_plugin stubs.

Thom1729:
Based on comments in https://github.com/SublimeText/sublime_lib/pull/160:

- Make `CommandInputHandler` generic.
- Offer overload for `confirm()` with event.
- Broaden type for `run()`.


gwenzek:
Looks good for me :-)

stubs/sublime_plugin.pyi
```diff
@@ -1,27 +1,35 @@
...
 
 
-class TextInputHandler(CommandInputHandler):
+class TextInputHandler(CommandInputHandler[str]):
     def description(self, text: str) -> str: ...
 
 
-class ListInputHandler(CommandInputHandler):
+class ListInputHandler(CommandInputHandler[InputType], Generic[InputType]):
     def list_items(self) -> list: ...
```
FichteFoll, line 33: Not required for us, most likely, but the return type of `list_items` is much more complex and needs to consider `InputType`. I don't know how detailed we want the stubs to be, though.

cc @jfcherng, since you didn't touch that in https://github.com/jfcherng-sublime/ST-API-stubs/commit/71950113a9473514590b4e720ea3e5886b467d3a#diff-d4b7ec302230de6d2cd37a74fe45305e8eb37f208cc9ab88b2f52301b99b3bd8R716

stubs/sublime_plugin.pyi
```diff
@@ -1,27 +1,35 @@
 import sublime
-from typing import Optional, Union, List, Dict, Tuple
 
+from typing import Callable, Generic, TypeVar, Optional, Union, List, Tuple, overload
 
-class CommandInputHandler():
+
+InputType = TypeVar('InputType')
```
FichteFoll, line 6: We could limit this to something like a `Union[str, int, float, list, dict, tuple, None]`, since it must be serializable, but optional for me.

Thom1729: Fixed.

stubs/sublime_plugin.pyi
Thom1729: Fixed.

Thom1729:
Other improvements:

- Remove the unnecessary inheritance clauses.
- `meta_info` returns `dict`, not `str`.
- `on_X_command` can return `None`.
- `on_query_context` returns `Optional[bool]`, not `Optional[None]`.
- Use more specific return type for `on_query_completions`.

FichteFoll (approved):

PR #168, opened 2021-05-03
Never use an existing enum module.

Thom1729:
See https://github.com/sublimelsp/LSP-typescript/issues/55.

In 1.5, we use an existing `enum` module if there is one. However, the `enum` dependency is a port of `enum34`, which does not support all of the features we need. This PR always uses the vendored `enum` module, but keeps the other import in unreachable code to keep MyPy happy.

If there are no issues, we'll want to push this out as 1.5.1 ASAP.


FichteFoll:
If you need a review urgently for me, don't hesitate to hmu on Discord. Unfortunately, I don't have the time currrently to check github notifications daily.

PR #172, opened 2021-05-21
Change typing stub to make pyright happy.

### LINKED ISSUE #171: Type annotations don't work correctly with pyright
Opening this project in ST with LSP-pyright enabled highlights all type annotations with errors. This is because pyright can't handle try/except based import logic like in:
https://github.com/SublimeText/sublime_lib/blob/54bcf41ee8de4a7c6482a3e60aa0d72383465d96/st3/sublime_lib/_compat/typing.py
It will always use the "last" import in such case.

The solution to that is doing a check based on python version like in: https://github.com/sublimelsp/LSP/blob/7fc1f80c3d3730b34d1f2fb8be65e41cfb43f2fe/plugin/core/typing.py#L3-L3

Fixing that would improve developer experience.

Thom1729:
Fix #171.

### DIFF
--- a/st3/sublime_lib/_compat/typing.py
+++ b/st3/sublime_lib/_compat/typing.py
@@ -1,4 +1,6 @@
-try:
+import sys
+
+if sys.version_info >= (3, 8, 0):
     from typing import *  # noqa: F401, F403
-except ImportError:
+else:
     from .typing_stubs import *  # type: ignore # noqa: F401, F403


FichteFoll (approved):

FichteFoll:
Do you want to make a hotfix release for this?

st3/sublime_lib/_compat/typing.py
```diff
@@ -1,4 +1,6 @@
-try:
+import sys
+
+if sys.version_info >= (3, 8, 0):
     from typing import *  # noqa: F401, F403
-except ImportError:
+else:
     from .typing_stubs import *  # type: ignore # noqa: F401, F403
```
rchl, line 6: Should it be something like this instead so that the Sublime Text `typing` depedency is used when installed?

```py
if sys.version_info >= (3, 8, 0):
    from typing import *  # noqa: F401, F403
else:
    try:
        from typing import *  # noqa: F401, F403
    except ImportError:
        from .typing_stubs import *  # type: ignore # noqa: F401, F403
```

Although I'm not sure what would be the point really. It would not change the runtime behavior and is not needed for pyright either.

rchl:
If you would want to ignore the `reportWildcardImportFromLibrary` warning you can add a configuration file in the root:

```diff
diff --git a/pyrightconfig.json b/pyrightconfig.json
new file mode 100644
index 0000000..d8bd1c1
--- /dev/null
+++ b/pyrightconfig.json
@@ -0,0 +1,4 @@
+{
+    "pythonVersion": "3.8",
+    "reportWildcardImportFromLibrary": "none"
+}
```

The `pythonVersion` is not relevant to the warning but would be good to have it explicit so that it guards against using "too new" features.

rchl:
Can we get this out?

FichteFoll:
@Thom1729 can you?

st3/sublime_lib/_compat/typing.py
Thom1729: Probably not needed, and we've had issues with a similar approach with `enum`.

FichteFoll:
Do we not need a (hotfix) relase for it?

Thom1729:
If we can merge #166, we can release 1.6 and there's probably no need for a hotfix. If not, I can release a hotfix tonight or tomorrow.

FichteFoll:
SGTM.

PR #178, opened 2025-11-21
Refactor repository to deploy as WHEEL via Github releases

deathaxe:
This PR...

1. refactors folder structure as if sublime_lib was an ordinary python library.
2. adds pyproject.toml to specify all relevant metadata
3. retires flake8 and pycodestyle in favour of black.
4. adds relevant scripts and actions to deploy the library as WHEEL via Github releases.

Version is bumbed to 1.6.0, but this PR doesn't change any library functionality at all.

PR #179, opened 2025-11-29
Structural updates, require Python 3.8, deploy a WHEEL file

FichteFoll:
This is an alternative approach to #178, using it as an inspiration but going about it slightly differently.

I tried to isolate commits into reasonable chunks.

Things I did:

- Update repository structure (rename `st3` to `src`).
- Add pyproject.toml file with dev dependencies & using `uv` for our toolchain.
- Require Python 3.8, since 3.3 is finally being deprecated by ST.
- Remove vendor and combat folders since they are not necessary in 3.8.

Things I did not do:

- Ensure that tests can still be run with Unittesting.
- Test anything besides `uv build`.
- Reformat or add a formatter configuration.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,7 +1,9 @@
 name: Publish Release
 
 on:
-  workflow_dispatch:
+  push:
+    tags:
+      - v*
 
 jobs:
   release:
@@ -23,4 +25,4 @@ jobs:
       - name: Create release & upload wheel
         run: |
           version=$(uv version --short)
-          gh release create --generate-notes --latest -t "v${version}" "v${version}" dist/*.whl
+          gh release create --generate-notes --latest -t "$GITHUB_REF_NAME" "$GITHUB_REF_NAME" dist/*.whl
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "sublime_lib"
-version = "1.6.0"
+dynamic = ["version"]
 description = "Utility library for frequently used functionality in Sublime Text"
 readme = "README.md"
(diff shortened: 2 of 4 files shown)


deathaxe:
> Require Python 3.8, since 3.3 is finally being deprecated by ST.

IMHO, we should first create a py33 compatible WHEEL release, targeting legacy ST3/4 builds, to totally get rid of legacy dependency releases without breaking unmaintained py33 packages on ST3/4 platform.

A 2nd step then would be to upgrade code base to a py38-only version.

pyproject.toml
```diff
@@ -0,0 +1,46 @@
...
+classifiers = [
+    "Intended Audience :: Developers",
+    "Operating System :: OS Independent",
+    "Programming Language :: Python :: 3",
+    "Topic :: Software Development :: Libraries :: Python Modules",
+]
+requires-python = ">=3.8,<=3.9"
+
+[build-system]
+requires = ["hatchling>=1.20", "hatch-vcs>=0.4"]
```
deathaxe: What's wrong with flit?

deathaxe:
For python 3.8 it may be useful to disable annotation evaluation via `from __future__ import annotations` to save some CPU cycles.

.github/workflows/release.yml
```diff
@@ -0,0 +1,28 @@
+name: Publish Release
+
+on:
+  push:
+    tags:
+      - v*
```
deathaxe: The idea with using dynamic versioning via pyproject.toml specifically is to avoid manual creation of version tags, but let them automatically be created from a single source of truth, the `_version.py`.

deathaxe:
Has all been applied via #178, but with py33 compat as a first step.

